Issue #9, opened 2026-05-11
ix switch --build-on local hangs silently after NAR upload completes


andrewgazelka:
After fixing the fleet to pass a system out-path (#8), `ix switch <vm> <out-path> --build-on local` for the minecraft-fleet lobby:

1. `nix-store --query --requisites` + `nix-store --export` runs fine locally.
2. `vm_fs_write_all_bytes` uploads ~2.4 GiB to lobby (matches the 2.3 GiB `nix path-info -S` of the system closure).
3. Network counters on the ix CLI process then sit unchanged for 30+ minutes. ix process is 4.84 GiB RSS, all tokio workers in `pthread_cond_wait`, no further activity.
4. On lobby: `nix-daemon` and `ix-agent` are alive, no `nix-store --import` process running, no `/tmp/ix-system-*.nar` present, `journalctl` shows no bash/script invocation at all between bootstrap finish and now. `/nix/store` has 6515 paths totalling ~1.2 GiB (partial closure), but the system out-path itself is missing. `/run/current-system` still points at the bootstrap.

Either the in-VM script (`nix-store --import < tmp.nar; rm; sys/bin/switch-to-configuration switch`) is never dispatched after the file write, or it ran and the SDK never observed its exit. Either way `ix switch` has no timeout, no progress indicator past upload, and no way to recover except SIGKILL. The result also indicates the `nix-store --import` step is not all-or-nothing here: 1.2 GiB landed in the store but the system path is absent, so `switch-to-configuration` could not run even if it had been dispatched.

Wanted:
1. Surface upload completion + script dispatch as observable events on the CLI side, with a timeout on the in-VM exec.
2. Verify that `vm_exec_to_stdio_checked` actually receives an exit status; if the channel can wedge, add a heartbeat or detect-and-retry.
3. Fail loudly when the closure import on the VM doesn't include the target path (e.g. `nix-store --verify-path $system` before invoking `switch-to-configuration`).

andrewgazelka:
Tracked in Linear ENG-2423; closing the GitHub issue as part of the issue sweep. The ix switch command appears to have been superseded by ix up (indexable-inc/ix#4442); the Linear ticket notes this and asks that it be verified before implementing the fix.

(triaged by an AI agent, Claude Fable 5)

PR #10, opened 2026-05-11
Add switch timeout and retry on stream framing errors

### LINKED ISSUE #6: ix-fleet should retry transient 'stream framing error' from ix switch
Running `nix run .#switch` for examples/minecraft-fleet, the remote build streamed `ix switch: remote nix build still running...` for ~2 minutes then died with `stream framing error: frame I/O failed: connection lost: reset by peer`. The lobby VM itself was healthy (no OOM, no kill, nix-daemon up), so the drop was somewhere between the dev box and ix's build orchestration. `ix-fleet.py` propagated the non-zero exit and stopped the whole fleet rollout.

Either `ix-fleet.py` should retry transient stream errors on the same build (the build target is content-addressed, so retry is cheap), or `ix switch` itself should reconnect to in-flight builds across stream drops. Right now any long znver5 build is one ECONNRESET away from a full rerun.

visualstudioblyat:
Fixes #6. Refs #9.

Adds timeout handling to `run_cli`, 30min/60min timeouts on switch commands, a retry loop for transient stream framing errors, and progress logging.

This is a fleet-side mitigation for #9 so stalled `ix switch` calls do not hang forever, but the lower-level switch dispatch/import verification work remains tracked there.

### DIFF
--- a/tools/ix-fleet.py
+++ b/tools/ix-fleet.py
@@ -4,8 +4,11 @@ from __future__ import annotations
 import argparse
 import asyncio
 import json
+import os
+import select
 import subprocess
 import sys
+import time
 import typing
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


andrewgazelka:
I reviewed this and pushed a follow-up to the PR branch because the original retry path likely would not have fired for the real failure: `run_cli` only captured stdout, while the `stream framing error` text comes from stderr, and `CalledProcessError.__str__()` does not include stderr.\n\nThe follow-up keeps stdout/stderr streamed to the terminal, retains them separately, returns stdout for JSON callers, and makes the retry check inspect the captured command output. I also rebased the PR onto current `main`.\n\nValidation I ran on the updated branch:\n- `nix run nixpkgs#ast-grep -- scan`\n- `nix flake check --no-build`\n- `nix build .#checks.x86_64-linux.eval --no-link`\n- `nix run .#ix-fleet -- --help`\n\nThis is now a reasonable fleet-side mitigation for #6 and gives #9 a timeout instead of an infinite hang. It still does not fully solve the lower-level #9 asks around in-VM exec heartbeat/script-dispatch visibility or verifying the imported system path before switching; those likely belong in `ix switch` itself.

PR #54, opened 2026-05-18
Fix #51: Make ix CLI a self-fetching package

### LINKED ISSUE #51: Mutable ix CLI URL breaks flake check when locked hash drifts
`nix flake check -L` currently fails before reaching repo checks because the locked `ixCliX86_64Linux` file input no longer matches the bytes served by `https://ix.dev/cli/linux-x86_64/ix`.

To reproduce:

```sh
nix flake check -L
```

Observed error:

```text
mismatch in field 'narHash' of input ... url="https://ix.dev/cli/linux-x86_64/ix"
```

Desired outcome: make the ix CLI flake inputs immutable/versioned, or update the locking workflow so routine flake checks do not fail when the URL target changes.

Paramount50:
Closes #51.

Following the rule in `AGENTS.md` ("No artifact URLs in flake.nix inputs"), this removes the mutable `ixCli*` flake inputs and instead fetches the precompiled binaries directly in `packages/ix/default.nix` using `pkgs.fetchurl` with locked hashes. 

This prevents `nix flake check` from crashing when the `ix.dev` latest URL drifts.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,41 +1,5 @@
 {
   "nodes": {
-    "ixCliAarch64Darwin": {
-      "flake": false,
-      "locked": {
-        "narHash": "sha256-lPxd7XkazzfTkM2Az4R1ENlEmXzaF70O93OagH1Kn2s=",
-        "type": "file",
-        "url": "https://ix.dev/cli/darwin-arm64/ix"
-      },
-      "original": {
-        "type": "file",
-        "url": "https://ix.dev/cli/darwin-arm64/ix"
-      }
-    },
-    "ixCliX86_64Darwin": {
-      "flake": false,
-      "locked": {
-        "narHash": "sha256-ivkOJr+NUsoeyZ5m90LcS1BC89ibnzVKPui+Z7GdjkQ=",
-        "type": "file",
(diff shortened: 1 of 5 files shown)


andrewgazelka:
Thanks for the PR. I am going to close this rather than merge it because this exposed that our AGENTS.md guidance was too broad for the ix CLI case.

For catalog/data artifacts, keeping URL + SRI next to the consumer still makes sense. For repo-owned workflow tools like the prebuilt ix CLI, making contributors manually refresh per-platform hashes is worse than using the lockfile as the update surface. The durable fix for #51 should be to make the ix CLI input immutable/versioned, or expose it through an upstream flake, so old locks keep building and routine bumps are just `nix flake update`.

I updated the repo guidance on `main` in 2d3a863 (`Clarify flake input policy for tool pins`) to capture that distinction. Closing this PR so the issue can be fixed in that direction instead.

andrewgazelka:
Closing per the maintainer note above. The replacement direction is immutable/versioned ix CLI inputs with `nix flake update` as the normal bump path, rather than moving these hashes into package code.

PR #64, opened 2026-05-19
feat: add nginx lifecycle fleet example

TestingPlant:
Adds an index-owned nginx lifecycle fleet example and teaches ix-fleet plans to mark nodes that should be recreated on each up. The example exposes nginx on port 80, declares the port claim, and includes guest health checks for systemd and HTTP loopback.

This moves the fresh VM launch probe into index so external schedulers can run it with the rest of the fleet integration suite. The recreate flag keeps the lifecycle semantics explicit inside the fleet plan.

### DIFF
--- a/examples/nginx-lifecycle/default.nix
+++ b/examples/nginx-lifecycle/default.nix
@@ -1,0 +1,10 @@
+{ index }:
+
+index.lib.mkFleet {
+  defaults = [ { ix.image.tag = "nginx-lifecycle"; } ];
+
+  nodes.nginx = {
+    deployment.recreateOnUp = true;
+    modules = [ ./service.nix ];
+  };
+}
--- a/examples/nginx-lifecycle/service.nix
+++ b/examples/nginx-lifecycle/service.nix
@@ -1,0 +1,46 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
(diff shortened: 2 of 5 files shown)


andrewgazelka:
Claude: small housekeeping pass on this PR.

- Re-targeted base from `main` to `development` per the `## Branch model` section of `AGENTS.md` ("PRs target `development`; routine work lands there after local checks and review"). `main` is reserved for the scheduled fast-forward promotion from `development`.
- Verified the branch rebases cleanly onto current `development` (no conflicts in `lib/fleet.nix`, `tests/default.nix`, or `packages/ix-fleet/src/ix_fleet/__init__.py`), so no force-push needed from your side beyond a normal rebase.
- The earlier `Check` failure on this branch was a `zlib` checkPhase `Illegal instruction` — the `znver5` host platform compiles binaries with AVX-512 / Zen 5 instructions, and the GitHub-hosted Azure runner (`eastus`) doesn't have those instructions, so any test-phase binary execution traps. Separately, `development` itself currently fails `nix flake check` with `error: attribute 'shell' missing` at `tests/default.nix:1004`, which a rebase would inherit — that's a tree-wide issue, not something for this PR to fix.

Did not force-push to the branch (your work, not mine to rewrite). Re-targeting + a rebase from your side should pick up everything.

PR #196, opened 2026-05-26
minecraft: preserve legacy manifest paths

andrewgazelka:
## Summary

Fixes the PR #139 review finding that legacy managed-file manifest entries with spaces are split as if the last word were a target path.

- treats a manifest target as present only when the last field is an absolute path
- keeps generated `rel /absolute/source` manifest entries working
- adds a unit assertion for a legacy path with spaces and no target

Review thread:

- https://github.com/indexable-inc/index/pull/139#discussion_r3303339318

## Checks

- `nix run nixpkgs#rustfmt -- --edition 2024 --check packages/minecraft/sync-managed/src/main.rs`
- `nix build .#minecraft-sync-managed --no-link`
- `nix run .#lint`


andrewgazelka:
@codex review

PR #199, opened 2026-05-26
cargo-unit: scope doctests to target sets

andrewgazelka:
## Summary
- render each target set's doctest map from that target set's roots
- keep the top-level doctest map unchanged
- add a regression fixture with two root sets whose doctests must stay isolated

## Checks
- `nix run nixpkgs#rustfmt -- --edition 2024 --check packages/nix-cargo-unit/src/render.rs`
- `nix build .#nix-cargo-unit.passthru.tests.package --no-link`
- `nix run .#lint`

Refs #143


andrewgazelka:
@codex review

PR #210, opened 2026-05-26
go-unit: reject unreadable no-sum modules

andrewgazelka:
## Summary
- require `vendorHash = null` to be backed by a readable `go.mod` with no `require` directives
- keep derivation-backed stdlib modules working when callers pass an explicit readable `goMod`
- add a regression for unreadable derivation sources requesting no-sum mode

## Checks
- `nix build .#checks.x86_64-linux.eval --no-link`
- `nix run .#lint`
- `git diff --check`

Refs #187 review thread: https://github.com/indexable-inc/index/pull/187#discussion_r3305000396


andrewgazelka:
@codex review

PR #212, opened 2026-05-26
modules: auto-discover from the filesystem (#1)

wyattgill9:
Replaces the hand-maintained registry in `modules/default.nix` with a filesystem walker (`lib.discoverModules`) that mirrors `discoverImages`. Adding a module now means dropping a `default.nix` under `modules/<category>/<name>/`; the next eval picks it up.

Every existing flat `<name>.nix` module is restructured into `<name>/default.nix` form so the walker only has to look for one filename. The walker skips any path segment starting with `_`, which is why `services/observability/dashboards/` moves to `_dashboards/` (it holds dashboard data helpers, not NixOS modules).

The `minecraft.runtime` key is renamed to `minecraft.default` to match the new convention; no internal or external consumers reference the old name.

@andrewgazelka


andrewgazelka (approved):
Approved current head after adding the missing AGENTS.md fragment source and running `nix run .#lint` locally.

PR #233, opened 2026-05-27
[codex] skip Cachix setup on PR checks

wyattgill9:
## Summary

- skip `cachix/cachix-action` on pull request `Check` runs
- keep Cachix publishing on push events for `main` and tags
- rely on the accepted flake `nixConfig` for PR read-only substitution

## Why

Recent timing showed normal PR checks spend about 24 seconds in Cachix setup, and one outlier spent about 17 minutes there due to repeated install/cache HTTP retries. PR jobs do not publish to Cachix because they do not receive `CACHIX_AUTH_TOKEN`, so the action is redundant for that path.

## Validation

- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(ARGV.fetch(0)); puts "yaml ok"' .github/workflows/check.yml`
- `nix run .#lint`


wyattgill9:
@andrewgazelka

PR #247, opened 2026-05-28
fix(cargo-unit): per-package [lints] resolution and clippy UX

andrewgazelka:
fix(cargo-unit): per-package [lints] resolution and clippy UX

Three follow-ups to per-unit clippy:

1. **Per-package `[lints]` resolution (the load-bearing fix).** Previously
   the workspace `[workspace.lints.clippy]` table was applied to every
   local unit, including path-vendored crates that never declared
   `[lints] workspace = true`. That diverged from `cargo clippy`, which
   only applies the workspace policy to crates that opt in via
   `[lints]`. Vendored crates ended up linted under your workspace
   policy and failed on upstream style (`unwrap()` etc.).

   `render.rs` now resolves clippy flags per unit at prepare time:

   - Read each unit's package `Cargo.toml`.
   - If `[lints] workspace = true`, inherit `[workspace.lints.clippy]`.
   - Else if `[lints.clippy]` is present, use those entries.
   - Else, emit no clippy flags (matches `cargo clippy` for a crate
     without a `[lints]` table — vendored upstream code is unaffected).

   Lint entries are sorted by ascending priority (so per-lint allows
   override group denies, matching cargo) and rendered as
   `-D|-W|-A|-F clippy::<name>`. `clippy::cargo` group members are
   filtered because they need `cargo metadata` access that the per-unit
   sandbox doesn't have.

   The nix-side `clippyLintFlagsFromManifest` helper in
   `lib/cargo-unit.nix` is removed; `extraClippyLintArgs` collapses to
   `rust.clippyLintArgs args.policy` as the nix-level policy escape
   hatch only.

2. **`pname` includes the package name.** Build-script compile units all
   share `target.name == "build-script-build"`, so the previous pname
   `build-script-build-clippy-<version>` was identical across every
   crate's build.rs failure. Now it's
   `<package>-build-script-build-clippy-<version>` (and
   `<package>-<target>-clippy-<version>` for everything else), so a
   failing unit identifies its owning crate at a glance.

3. **Clippy diagnostic survives nix's default log tail.** `nix build`
   truncates a failed derivation's log to the last ~25 lines, and bash
   stdenv cleanup runs before exit, so the actual clippy error gets
   pushed off-screen by failure hooks. The clippy build phase now
   captures driver stdout/stderr to `build/clippy-driver.log`, prints
   it as the run produces it, and re-tails the last 80 lines after a
   non-zero exit so the clippy diagnostic shows up in the default tail.
   `nix build -L` still streams the live log.

### DIFF
--- a/lib/cargo-unit.nix
+++ b/lib/cargo-unit.nix
@@ -66,72 +66,6 @@ let
       Fetched or patched sources pass workspaceRoot = src.
     '');
 
-  # Cargo only emits `[lints.clippy]` into the unit graph's `lint_rustflags`
-  # when invoked as `cargo clippy`, not `cargo build`. Our unit graph is built
-  # with `cargo build --unit-graph`, so per-unit clippy never sees the
-  # workspace lint policy unless we resolve it ourselves. Parse the workspace
-  # manifest and emit the equivalent `-D|-W|-A clippy::<lint>` flags.
-  #
-  # Per-package overrides (a package with its own `[lints.clippy]` and
-  # `workspace = false`) are not yet honored; most workspaces in practice use
-  # `[lints] workspace = true` per crate, which inherits the workspace table.
-  #
-  # `clippy::cargo` group lints and the individual members of that group
-  # invoke the `cargo` binary to read workspace metadata. Per-unit clippy
-  # runs in a sandboxed build directory without a discoverable Cargo.toml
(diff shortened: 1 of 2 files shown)


andrewgazelka:
i have no diea what it is doing here lol

PR #248, opened 2026-05-28
cargo-unit: add per-target miri checks

andrewgazelka:
## Summary

- add per-target `miriUnits` derivations to rendered cargo-unit workspaces
- add `makeMiriTest` / `miriTests` as aggregate surfaces over selected Miri units
- cover the shape with renderer assertions and an eval fixture that proves a named Miri preset stays a join over a target derivation

## Checks

- `nix shell nixpkgs#cargo nixpkgs#rustc nixpkgs#stdenv.cc -c cargo test -p nix-cargo-unit exposes_test_roots_as_runnable_checks`
- `nix eval .#checks.x86_64-linux.eval --json`
- `nix build .#checks.x86_64-linux.eval`
- `nix run .#lint`

### DIFF
--- a/lib/cargo-unit.nix
+++ b/lib/cargo-unit.nix
@@ -38,6 +38,7 @@ let
     cargoExtraConfig = args.cargoExtraConfig or "";
     vendorDir = args.vendorDir or null;
     vendorSources = args.vendorSources or null;
+    miriTests = args.miriTests or { };
     # Maps exact Cargo.lock source strings to already-fetched source trees.
     # This keeps private Git dependencies reproducible without requiring
     # sandboxed fetchers to see a developer SSH agent or GitHub credentials.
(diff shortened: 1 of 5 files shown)


andrewgazelka:
Backlogging this Miri surface for now.

Current state proves the rough fan-out shape, but it is not ready to merge. The hard part is the source/workspace boundary for `cargo miri test`:

- Miri cannot run the already-built native cargo-unit test binaries, so the practical path is still `cargo miri test` per test target.
- Running Cargo from the immutable Nix store source is wrong because tests can write relative to cwd.
- Copying `workspaceRoot` gives a writable full workspace, but can diverge from the filtered/patched `src` that produced the unit graph.
- Copying `src` preserves unit-graph consistency, but may be incomplete for callers that intentionally use a narrow graph source plus a broader `workspaceRoot` for real package source closures.

TODO: design a proper Miri source model before making this mergeable. Likely this needs an explicit Miri workspace source/copy phase, or a checked restriction that rejects filtered graph sources for Miri presets with a clear error. Until then this PR stays draft.

andrewgazelka:
Closing for now; will revisit. Branch kept.

PR #271, opened 2026-05-28
symphony: build room server with static crates

harivansh-afk:
Build the baked Symphony room server through the index Rust builder instead of importing Symphony's raw flake package directly. The index builder vendors crates through the static crate CDN path, so CI and image builds avoid the crates.io API endpoint that returns 403 on GitHub runners.

Validation:
- `nix run nixpkgs#nixfmt-rfc-style -- lib/default.nix lib/overlay.nix`
- `nix --accept-flake-config --option eval-cache false eval --raw .#lib.pkgs.symphony-room-server.name`
- `nix build --accept-flake-config .#lib.pkgs.symphony-room-server --no-link --dry-run`

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -44,8 +44,10 @@ let
     inherit
       lib
       packageRegistry
+      rust-overlay
       symphony
       buildIxRustTool
+      buildRustPackage
       clippy-fork
       writePythonApplication
       ;
--- a/lib/overlay.nix
+++ b/lib/overlay.nix
@@ -1,14 +1,49 @@
 {
   lib,
   packageRegistry,
+  rust-overlay,
   symphony,
   buildIxRustTool,
+  buildRustPackage,
   clippy-fork,
   writePythonApplication,
 }:
 final: _prev:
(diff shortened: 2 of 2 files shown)


andrewgazelka:
Closing in favor of fixing this at the source in symphony's own flake.

The crates.io 403 on GitHub runners is real and the static-crate-CDN vendoring is the right mechanism. The objection is ownership: this re-derives `room-server` inside index's overlay with a local `buildRustPackage`, which drops symphony's own build definition and makes index a second owner of how symphony builds. index should keep consuming `symphony.packages.${system}.room-server`; the static-crate-CDN path belongs in symphony's flake so the fix lands once at the owner and every consumer benefits.

Ready-to-run for an agent or follow-up PR, in the symphony repo: switch the `room-server` package to vendor crates through the static crate CDN (the same `Cargo.nix.lock` + CDN approach used here), then drop this overlay change so index returns to a thin passthrough.

Thanks @harivansh-afk.

— via Claude (Opus 4.8)

Issue #307, opened 2026-05-29
cargo-unit: per-unit panic-freedom check via compiled-artifact scan


andrewgazelka:
### Problem

cargo-unit renders each Cargo unit-graph node into its own isolated Nix derivation, but there is no way to assert that a unit's compiled code cannot panic. We want an opt-in per-unit check that inspects the compiled artifact and fails the build if the crate can panic, so panic-freedom becomes a checked property rather than a hope.

This is distinct from the existing `Panic` profile enum in `packages/nix-cargo-unit/src/model.rs` (the codegen `panic=abort/unwind` strategy). That picks the unwinding mechanism; it does not prove the code never reaches a panic. Lint-level gates (`clippy::panic`, `unwrap_used`, ...) are explicitly out of scope: they are a style gate that misses implicit panics in dependencies. This check works at the artifact tier only.

### Desired outcome

An opt-in flag (e.g. `--deny-panics`, or a per-unit attribute) that attaches a panic-freedom check to the generated unit derivation. The check inspects the unit's compiled rlib/object for references to `core::panicking::panic*` landing pads, so it sees panics in first-party code and dependencies alike. Because each unit is already its own derivation, the scan runs per-unit and caches like any other build output. A unit that can panic fails with an operator-facing message naming the offending function/site, not a backtrace.

### Known failure modes to handle

- **False positives:** panics the optimizer proves unreachable but does not strip. The check must report the symbol/site so an operator can see what survived.
- **Opt-level sensitivity:** panic visibility depends on the profile the unit was built with. The check reads whatever the unit actually compiled, and the message should state the profile so a debug-vs-release discrepancy is legible.

### Push back before building (per AGENTS.md dependency-intake rule)

Prefer wrapping an existing artifact scanner over hand-rolling symbol inspection:

- `findpanics` / `cargo-nopanic`: scan a built binary/rlib for panic calls.
- `no-panic` (dtolnay): attribute macro, link error if a function can panic. Function-level, needs annotations, fragile across opt levels.

The concrete gap that may justify ownership: these target whole binaries, while cargo-unit owns the per-unit rlib boundary and can run the scan at exactly that granularity with caching. Name that gap explicitly before reimplementing; if `findpanics` can be pointed at a single rlib, wrap it.

### Acceptance

- [ ] Decision recorded: wrap `findpanics`/`cargo-nopanic` or own the rlib scan, with the per-unit gap named.
[...]

PR #309, opened 2026-05-29
cargo-unit: per-unit panic-freedom check via relocation scan

andrewgazelka:
Adds an opt-in `--deny-panics` policy check to cargo-unit that fails a build when a workspace function can reach a panic. Refs #307.

## Approach

Detection is relocation-based. rustc lowers a panic call to a relocation whose target is an undefined panic sink (`core::panicking::*`, plus the `unwrap_failed` / `expect_failed` cold helpers that `unwrap`/`expect` route through, confirmed against the relocations rustc emits at opt-level 0 and 3). The new `scan-panics` subcommand reads symbols and relocations with the `object` crate and attributes each panic call to its containing function, with no disassembler and no demangler, so one path covers ELF and Mach-O.

To cover generics, the scan reads the relocatable objects of every workspace unit, not just library rlibs. A generic carries no relocation in its defining rlib because it is codegened where it is monomorphized; the bin or test object that instantiates it does carry the relocation, and the symbol keeps the defining crate's mangled token. Each unit therefore gets a `panic-objects` derivation (a recompile with `--emit obj`, mirroring the clippy units), and findings are scoped to the whole workspace crate set so a library generic monomorphized inside another unit is attributed back to its crate.

This was the deliberate design after ruling out the naive alternative: a trivial clean std binary already carries ~36 `panicking` symbols from std's runtime, so a whole-artifact symbol grep is always-red for std code. Relocation attribution scoped to the workspace crate set is the signal that actually means "this workspace's code can panic."

## What it does

- `--deny-panics` renders `policyChecks.panicFreedom`: one scan derivation per workspace unit, joined under one aggregate, so a touched unit only re-scans itself.
- Each scan runs `nix-cargo-unit scan-panics --crate-name ...` over the unit's `panic-objects` derivation, scoped to the workspace crate set.
- The scanner is the `cargoUnit` package, threaded through a new `cargoUnit` template argument and asserted non-null, so enabling the policy without wiring the scanner fails loudly instead of silently passing.

## Validation

- 45 Rust tests pass (hermetic scanner tests synthesize objects via `object::write`, no rustc dependency), clippy clean, fmt clean, generated Nix parses.
- End-to-end on real crates: `unwrap`/`expect`/indexing are flagged with their exact entrypoints; a generic `pub fn first<T>(xs:&[T])->T{xs[3]}` invisible in its own rlib is caught via a consumer's objects and attributed to its defining crate; a clean crate and a crate with its own `panicking` module both pass.

## Known limitations (documented in code)

- Best-effort, not a soundness proof. The residual gap is a public generic that no workspace bin/test ever instantiates (never codegened anywhere here, and also unreachable from the workspace's own entrypoints). The sound successor is call-graph reachability over the linked, monomorphized binary (what `findpanics` does).
- The panic-sink catalog is curated; a panic routed through some other std/alloc cold path is missed until its symbol is added.
- The `ix.cargoUnit.buildWorkspace` `policy` toggle that passes `cargoUnit` and sets `--deny-panics` lives in the `ix` repo and is the remaining integration step before this runs in a real workspace.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2311,6 +2311,7 @@ dependencies = [
  "askama",
  "clap",
  "color-eyre",
+ "object",
  "serde",
  "serde_json",
  "sha2 0.10.9",
@@ -2473,6 +2474,9 @@ version = "0.37.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ff76201f031d8863c38aa7f905eca4f53abbfa15f609db4277d44cd8938f33fe"
 dependencies = [
+ "crc32fast",
+ "hashbrown 0.15.5",
+ "indexmap",
  "memchr",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,11 @@ loro = "1.12"
 ndarray = "0.16"
 nix-web-monitor-parser = { path = "packages/nix-web-monitor/parser" }
 numpy = "0.28"
+object = { version = "0.37", default-features = false, features = [
+  "read",
+  "write",
+  "std",
+] }
 parking_lot = "0.12"
 pty-process = { version = "0.4", features = ["async"] }
 pyo3 = { version = "0.28", features = ["abi3-py311"] }
--- a/lib/cargo-unit.nix
+++ b/lib/cargo-unit.nix
@@ -256,7 +256,8 @@ let
         toolchainId
       ]
       ++ lib.optional args.contentAddressed "--content-addressed"
-      ++ lib.optional args.policy.denyUnusedCrateDependencies "--deny-unused-crate-dependencies";
+      ++ lib.optional args.policy.denyUnusedCrateDependencies "--deny-unused-crate-dependencies"
+      ++ lib.optional args.policy.denyPanics "--deny-panics";
     in
     pkgs.runCommand "cargo-units.nix"
       {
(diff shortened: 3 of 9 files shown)


andrewgazelka:
Claude: thanks for the reviews — they caught real issues. Status of each finding across the follow-up commits:

**Fixed in code**
- Missed `unwrap`/`expect` (route through `core::{option,result}::{unwrap,expect}_failed`, not `core::panicking`): curated panic-sink catalog. `b759bbe`
- Panic-sink matching too broad / not at the core boundary: `is_panic_entrypoint` now requires `core`/`std` as the crate root (`_`/`_ZN` boundary). `6ee75e7`
- Generics not codegened in the defining rlib: scan the per-unit `--emit obj` objects of every unit (lib/bin/test/bench) where monomorphization happens, scoped to the workspace crate set. `a7d95d8`
- `cdylib`/`staticlib`/`dylib` and rlib-only assumption: we now scan objects, not rlibs, so crate type no longer gates coverage. `a7d95d8`
- Unmangled `#[no_mangle]` exports dropped by the filter: included via `is_rust_mangled`. `b759bbe`
- Policy unreachable from `buildWorkspace`: added `policy.denyPanics`, forwarded `--deny-panics` from `lib/cargo-unit.nix`, and passed the `cargoUnit` scanner. `6ee75e7`
- Scan passes with nothing to scan / unparseable artifact ignored: `scan-panics` now fails closed on an empty artifact set and on any artifact it cannot parse (non-object archive members like rmeta are still skipped). `6ee75e7`
- Short crate token substring-matching dependency symbols (`de` vs `serde2de`): the workspace filter now matches tokens only at the crate-root boundary. `4b28abe`

**Documented residuals (best-effort by design, opt-in, off by default)**
- A public generic that no workspace bin/test/bench ever instantiates is never codegened anywhere here, so it is not scanned. It is also unreachable from the workspace's own entrypoints.
- A dependency generic instantiated over a workspace type can still be attributed to the workspace (the type argument carries the workspace crate at a boundary). This over-approximates in the gate-safe direction: it flags a panic the workspace binary actually reaches, attributed to the dep's function. Precise defining-crate-only attribution would require parsing the v0 symbol structure.

The sound successor to all of this is call-graph reachability over the linked, monomorphized binary (what `findpanics` does); noted in the module docs.

Issue #322, opened 2026-05-29
tui: share one dashboard across multiple processes via a unix-socket server


andrewgazelka:
## Problem

Each `tui-py` process runs its own in-process `serve()` dashboard. When several agents run at once (for example multiple Claude sessions), they each call `serve()`, collide on the dashboard port, and each page only shows that one process's terminals. There is no unified view across processes.

## Context

`serve()` today mirrors every `Tui` in the current process into one Loro CRDT doc and streams it to browsers over SSE (see `packages/tui` README "Web dashboard" and `packages/tui-py` `serve()`). The `TuiManager` owns the PTY actors in-process, so the dashboard's scope is one process.

## Idea to evaluate

Run a single dashboard server (a manually started daemon). Each `tui` process connects to it over a unix socket and syncs its terminals, e.g. Loro updates over the socket. Result: one server, one port, one aggregated grid spanning every process.

## Open design questions

- **PTY ownership.** Daemon aggregates viewport state while PTYs stay in each producer process, vs daemon owns the PTYs and producers send spawn/write commands. Aggregator is the smaller delta over today's `serve()`; a PTY-owning daemon would let terminals survive producer restarts.
- **Wire protocol.** Loro end-to-end (producer doc → socket → server doc → browser, server as pure relay/merge), vs simpler framed viewport diffs with the server building the Loro doc. Either way terminal ids must be namespaced per producer (pid) so concurrent writers in the shared doc never collide.
- **Discovery and lifecycle.** Well-known socket path (e.g. `$XDG_RUNTIME_DIR/ix-tui.sock`); if it is absent, return a typed error rather than silently falling back to a per-process `serve()` (repo "reject fallback paths" rule); is opt-in auto-spawn worth it? On producer disconnect, mark its terminals exited/dimmed via socket-close detection or a presence channel.
- **Alternative to a manual daemon.** First-process-wins: the first `serve()` binds the socket and port and becomes the server, later processes detect it and connect as clients. Removes the manual step but needs leader handoff when the holder dies.

## Desired outcome

A short design decision on aggregator-vs-PTY-owner and the wire protocol, then an implementation that lets multiple processes share one dashboard without port conflicts. Unix and macOS only; Windows is out of scope (PTY-backed, like the rest of `tui`).

andrewgazelka:
[...]

PR #326, opened 2026-05-29
tui: multi-process dashboard (unix-socket producer + aggregator) and UI redesign

### LINKED ISSUE #323: tui: publish a manager's terminals over a unix socket in a discovery dir
Part of #322 (sharing one dashboard across many processes). This is the **producer** half.

## Problem

`serve()` only renders terminals from its own process and binds its own HTTP port, so multiple agents cannot share one dashboard. We want each process to *publish* its terminals and let an external aggregator render them, with no HTTP server in the producer.

## Proposed shape

Add a producer mode where a `TuiManager` binds a unix socket under a discovery directory and streams its terminals to whoever connects:

- Socket path: `$XDG_RUNTIME_DIR/ix-tui/<pid>-<uuid>.sock` (the per-process uuid survives pid reuse and namespaces terminal ids in the aggregated view). Keep the directory short: macOS caps `sun_path` at 104 bytes.
- On connect, send a `hello` (producer id + current terminal list), then per-terminal viewport snapshots followed by diffs, plus lifecycle events (spawn, exit).
- Unlink the socket on clean shutdown; unlink a stale self-named socket on start.
- Directory `0700`, sockets `0600`: local-user only.

## Recommended call on the wire format

Producer emits **plain length-prefixed frames** (viewport rows + styled cells + lifecycle), not Loro. Each producer only ever writes its own terminals, so there is no concurrent-write conflict to merge on this hop, and keeping Loro out of the producer means every agent process stays dependency-light. Loro lives only on the aggregator → browser hop (see the aggregator issue).

## Context

Today's in-process Loro mirror + SSE is `tui::serve` in `packages/tui` (behind the `dashboard` feature), re-exported by `tui-py` and `tui-node`. This issue adds the publish path next to it.

## Desired outcome

A `tui` API (and `tui-py`/`tui-node` surface) that publishes a process's terminals to a discovery socket with no HTTP server of its own. Unix and macOS only.

### LINKED ISSUE #324: tui-dashboard: standalone aggregator that discovers terminal sockets and serves one grid
Part of #322 (sharing one dashboard across many processes). This is the **aggregator** half.

## Problem

We want one dashboard spanning every agent process, with no agent owning the server. A separate, independently-run process should discover the per-process sockets and present one combined grid.

## Proposed shape

A new standalone binary, run by a human (not launched by `tui`), e.g. `nix run .#tui-dashboard`:

- Scans the discovery directory for producer sockets and watches it for add/remove (kqueue on macOS, inotify on Linux).
- Connects to each producer, ingests its frame stream, and builds one Loro doc with one container per producer (id = pid+uuid). Serves the existing web grid over HTTP + SSE on a single port.
- On producer disconnect or connection-refused, mark that producer's terminals `exited`/dimmed, then evict after a grace period and reap the stale socket file.
- Exactly one process binds a TCP port (127.0.0.1 by default; `--host 0.0.0.0` to expose).

## Why this shape

- Pure discovery: drop a socket in the dir and you appear. Follows the repo's "discovery beats hand-maintained registries."
- The dashboard outlives any single agent, and no agent has to "win" a leader election.
- Reuses the dashboard HTML/SSE/Loro path already in `packages/tui`; only the data source changes from in-process to many sockets.

## Risks to handle

- Stale socket files after a hard crash: treat connect-refused as dead and reap.
- Loro doc growth across many short-lived terminals: evict exited producers and compact.

## Desired outcome

Run the aggregator once and watch every agent's terminals live, with producers coming and going freely, on a single stable URL. Unix and macOS only.

andrewgazelka:
## What

Lets many `tui` processes share one web dashboard instead of each starting its own server (the problem in #322). Splits the dashboard into a producer half and a viewer half:

- **Producer** (`tui::publish`, `publish` feature; `tui.publish()` in Python) binds a unix socket in a discovery directory and streams its manager's terminals as one NDJSON `ProducerSnapshot` per tick. No HTTP or CRDT dependency on the producer side, so a process that only publishes stays lightweight. Closes #323.
- **Aggregator** (`tui-dashboard` binary, run by hand) scans the discovery directory, connects to every producer socket, folds each producer's stream into one Loro document under its own scope, and serves the grid. No producer owns the server and exactly one process binds a TCP port, so agents come and go behind one stable URL. Closes #324.

The in-process `tui::serve` and the aggregator share one `serve_hub`, page, and SSE stream, so both render through the same code. The `dashboard` module is split into a scope-aware `hub` (the Loro doc + fan-out) and a `server` (router + SSE), and `serve` becomes a thin poller over a hub.

Also redesigns the dashboard web UI: a minimal, Ghostty-style dark grid of terminal cards (hairline header, status LED, dim producer chip, live terminal/producer counts).

## Design notes

- Producers emit plain framed snapshots; Loro lives only on the aggregator → browser hop. Each producer writes only its own terminals, so there is nothing to merge on the producer hop and every agent process stays Loro-free.
- Pure directory discovery: drop a socket in the dir and you appear; the aggregator reaps on disconnect or connection-refused. The discovery dir resolves to `$IX_TUI_DIR`, then `$XDG_RUNTIME_DIR/ix-tui`, then `/tmp/ix-tui-<user>`, kept short for the macOS 104-byte socket-path limit.
- Terminals are keyed `"<producer>\x1f<id>"`; reconcile and removal are scoped, so one producer never clobbers another.

## Tests

- `packages/tui/src/dashboard/hub.rs`: scope isolation (one producer's reconcile/disconnect never touches another's terminals), delta-only-on-change, empty-scope no-op.
- `packages/tui/tests/aggregate.rs`: a real `cat` PTY's echoed screen reaches a reader over the socket as a `ProducerSnapshot`, and `stop()` unlinks the socket.

## How to run

```sh
nix run .#tui-dashboard            # serve http://127.0.0.1:8080/ for the discovery dir
nix run .#tui-dashboard -- --help  # --host, --port, --dir, --rescan-ms
```

Refs #322.


andrewgazelka:
Thanks — addressed all of these in code:

- **chmod arbitrary parents** (`publish`): now only `restrict_dir` a directory this producer actually creates (guarded on `parent.exists()` before `create_dir_all`); a caller-supplied existing dir is left untouched. (`9613213`)
- **unlink non-socket before bind** (`publish`): replaced the unconditional `remove_file` with `reap_stale_socket`, which removes only an actual socket and returns a typed error for any non-socket path, so a typo never deletes data. Added unit tests. (`9613213`)
- **serve tasks on temporary runtimes**: re-added `TuiManager::runtime_handle()`; `serve` and `publish` now spawn the server, poller, and accept loops on the manager's long-lived runtime, so a sync `Runtime::new().block_on(serve(..))` survives the temp runtime dropping. (`9613213`)
- **aggregator reaping non-socket `*.sock`**: the `ConnectionRefused` reap in `read_producer` is now guarded by an `is_socket` check, so a regular file named `*.sock` in the watched dir is never deleted. (`c7d31e1`)

PR #335, opened 2026-05-30
fix(tap): rename integration test to avoid cross-package target-name collision

andrewgazelka:
## Problem

`main` is red: `.#checks.x86_64-linux.eval` fails the assertion in `tests/default.nix` that every repo Rust package with integration tests exposes an `integration-all` target, because `dag-runner.passthru.tests` lost its `integration-all` entry.

## Root cause

`nix-cargo-unit` disambiguates root test-unit keys by counting same-named cargo targets across the whole workspace, not per package (`compute_root_keys`). When #329 added `packages/tap/tests/integration.rs`, two packages had a root test target named `integration` (tap and dag-runner). The global count became 2, so both aggregate keys were hash-disambiguated: dag-runner's `integration-all` became `integration-0.1.0-<hash>-all`, dropping the `integration-all` attribute the assertion requires.

The integration test was always discovered; only its generated aggregate name was mangled by the cross-package collision.

## Fix

Rename `packages/tap/tests/integration.rs` to `packages/tap/tests/session.rs` (it drives a real PTY session: attach, resize, share), plus its one README link. This is a pure file rename; the test body is unchanged. Every other package names its integration test after what it covers (`nbt/tests/property.rs`, `file-search/tests/roundtrip.rs`, `tui/tests/aggregate.rs`), so the generic `integration.rs` was the anomaly. dag-runner's name is pinned by the assertion, so tap (the newcomer that introduced the clash) is the right side to rename. The assertion is not weakened.

## Known limitation

This fixes the colliding filename, not the latent `nix-cargo-unit` behavior: any future third package adding a duplicate root test-target name would re-mangle all sharers. A consumer-facing per-package target name independent of the global manifest key would be the deeper fix, but it would rename every package's `passthru.tests` attributes and is disproportionate to this regression. Worth a follow-up issue.


andrewgazelka:
Superseded by #333. The standalone dag-runner fix cannot pass CI on its own: main's old single-threaded `nix flake check --no-build` schema step also fails to realize the `elevenlabs-say` IFD source (`path '…-source' is not valid`), so this PR's flake-check died there even after the tap rename fixed the dag-runner eval check. #333 replaces that schema step with nix-eval-jobs (which realizes IFD on demand) and carries the same tap rename, so it fixes both in one green run.

PR #389, opened 2026-05-31
feat(lib): portable user-service module (launchd + systemd from one spec)

andrewgazelka:
## What

Adds a home-manager module, exported as `homeModules.portable-services`, that turns one declarative spec into a native **launchd agent** on macOS and native **systemd user units** on Linux:

```nix
services.portable.pr-watch = {
  command = [ "/run/current-system/sw/bin/pr-watch" "--once" ];
  interval = 300;            # every 5 min: StartInterval on launchd, a .timer on systemd
  restart = "on-failure";
  standardOutPath = "/tmp/pr-watch.log";
};
```

home-manager is the only module system that provides both `launchd.agents` (macOS, per-user `~/Library/LaunchAgents`) and `systemd.user.services` (Linux) from one config, so it is the natural base for a cross-OS module. nix-darwin only covers the macOS half (and uses the different `launchd.daemons.<name>.serviceConfig` tree), so it can't span both.

## Why this shape

It synthesizes the three existing approaches rather than picking one:

- **home-manager dual-schema** — emit *native* units per OS (real plist, real unit), not a foreign supervisor. Kept, but you write one spec instead of two hand-synced schemas.
- **services-flake** — write-once, cross-platform. Kept; the process-compose supervisor is dropped so units integrate with login/boot, journald, and restart directly.
- **RFC-163 modular services** — a portable option core with typed escape hatches. Kept: `launchd.config` and `systemd.service` accept raw keys deep-merged *over* the generated unit, so no power is lost when the portable subset isn't enough.

The portable subset (`command`, `environment`, `workingDirectory`, `runAtLoad`, `restart`, `interval`, `standardOutPath`/`standardErrorPath`) is exactly the set that maps cleanly onto both init systems. Restart maps `no` / `on-failure` / `always` to launchd `KeepAlive` and systemd `Restart=`; `interval` becomes launchd `StartInterval` or a systemd oneshot + `.timer`.

Pure transforms (`toLaunchdConfig` / `toSystemdUnits`) live in `lib/portable-services.nix` and are golden-tested in `tests/portable-services.nix` (wired into the `eval` check).

## Known limitations (does not yet cover every case)

- **User services only.** No system-level `launchd.daemons` / `systemd.services` (root) yet.
- **Interval, not calendar.** Periodic scheduling is `interval` (every N seconds). Calendar schedules (`StartCalendarInterval` / `OnCalendar`) are escape-hatch only for now.
- **systemd `ExecStart` quoting.** Argv is rendered with systemd's own quoting (not shell), but argv elements containing `%` hit systemd specifier expansion and aren't represented faithfully — use the `systemd.service` escape hatch for those.
- **No socket/path activation, no dependency ordering** in the portable core (escape hatches cover `Unit.After` etc).
- The portable↔native mapping is best-effort; the escape hatches are the intended pressure valve when a knob doesn't generalize.

Flagging @wyattgill9 to sanity-check the launchd/systemd semantic mappings (restart, interval, quoting) and whether the portable subset is the right cut — see the linked issue.

---
Made with Claude (Opus 4.8).

> [!NOTE]
> ### Add portable user-service home-manager module targeting launchd and systemd from one spec
> - Adds [lib/portable-services.nix](https://github.com/indexable-inc/index/pull/389/files#diff-062a2f28f34f916ca6ca914b56ea9346d3fa3e96c06d05e2b510576ab2843ad3) which defines a `serviceSubmodule` schema for declaring user services once, with renderers `toLaunchdConfig` and `toSystemdUnits` for macOS and Linux respectively.
> - The `homeModule` reads `options.services.portable` declarations and maps them to native `launchd.agents` on Darwin or `systemd.user.services`/`systemd.user.timers` on Linux using `lib.mkIf`.
> - Interval-based services produce a paired `.timer` unit on systemd and use `StartInterval` on launchd; restart policies, environment variables, working directories, and log paths are mapped across both platforms.
> - Platform-specific escape hatches (`launchd.config`, `systemd.service`) are deep-merged last via `lib.recursiveUpdate` to allow overriding generated fields.
> - Exposes `ix.portableServices` from [lib/default.nix](https://github.com/indexable-inc/index/pull/389/files#diff-84f980eccc1d3f99c10f8b0e6c5c5fc2af2069517533c4c4d73264a190e85188) and `homeModules.portable-services` from [flake.nix](https://github.com/indexable-inc/index/pull/389/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0); eval tests in [tests/portable-services.nix](https://github.com/indexable-inc/index/pull/389/files#diff-38a10fc78b6d06930821c881165083a22b5c8cfcc69fe9dff15b2554520d2416) validate quoting, timer wiring, and cross-platform field mapping.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9fd5667.</sup>
> 
>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -144,6 +144,9 @@
     {
       lib = ix;
       inherit (ix) nixosModules;
+      # Workstation-facing home-manager module: declare a service once, get a
+      # native launchd agent on macOS and native systemd user units on Linux.
+      homeModules.portable-services = ix.portableServices.homeModule;
       overlays.default = ix.overlay;
       packages = collect "packages";
       checks = collect "checks";
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -62,6 +62,12 @@ let
   # `discoverImages` for symmetry; see its doc-comment for the discovery rules.
   nixosModules = discoverModules { root = paths.modules; };
 
+  # Portable user-service layer (launchd + systemd from one spec). Lives
+  # outside `modules/` on purpose: it is a home-manager module, not a NixOS
+  # module, so it must not be swept into `nixosModules` above. Exposed to
+  # consumers as `homeModules.portable-services` from the flake.
+  portableServices = import ./portable-services.nix { inherit lib; };
+
   # Flat list of module paths from the auto-discovered registry under
   # `modules/`. Pulled in unconditionally so every option is in scope; each
   # module stays inert until its `enable` flag is set.
@@ -383,6 +389,7 @@ let
       discoverImages
       discoverModules
       nixosModules
+      portableServices
       exampleFleetsFor
       artifacts
       agentsMd
--- a/lib/portable-services.nix
+++ b/lib/portable-services.nix
@@ -1,0 +1,322 @@
+# Portable user services: one declarative spec that renders to a native
+# launchd agent on macOS and native systemd user units on Linux.
+#
+# Design (synthesis of the three existing approaches):
+#
+#   * home-manager dual-schema: produce *native* init units on each OS
+#     (a real launchd plist, a real systemd unit) rather than a foreign
+#     supervisor. We keep that, but the caller writes one spec instead of
+#     two hand-synced schemas.
+#   * services-flake: one definition, cross-platform. We keep "write once",
+#     but skip the process-compose supervisor: native units integrate with
+#     the OS (boot/login, logging, restart) without a long-lived parent.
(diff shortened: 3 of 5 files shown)


andrewgazelka:
Review tracked in #390 (assigned to @wyattgill9).

PR #497, opened 2026-05-31
indexbench: continuous-benchmarking framework

andrewgazelka:
## What

A durable, metric-centric continuous-benchmarking framework for the org. The core abstraction is a **Metric** (`name` / `value` / `unit` / `lower_is_better` / `samples`), not a time — so wall-clock, peak RSS, allocation counts, and arbitrary custom values all flow through one schema, one history store, and one comparator. New kinds of measurement never touch the schema.

```rust
Metric { name: String, value: f64, unit: String, lower_is_better: bool, samples: Option<Vec<f64>> }
Run    { suite, bench, metrics: Vec<Metric>, machine_id, git_commit, git_dirty, timestamp_unix }
```

## Components

- **Micro harness** (`micro`): warm-up + batched timed-loop sampler for a Rust closure (distributional `wall_clock`), plus a counting `#[global_allocator]` shim for a deterministic `allocations` count.
- **Macro harness** (`macro_harness`): runs an external command N times, reading `wall_clock` and `max_rss` from `libc::wait4`'s **per-child** `rusage` (avoids the `getrusage(RUSAGE_CHILDREN)` accumulation bug), and ingests the command's own structured lines as custom metrics — the extensibility hook:
  ```
  @bench name=<id> value=<f64> unit=<str> lower_is_better=<bool>
  ```
- **History store** (`store`): a pluggable `HistoryStore` trait. Default `GitBranchStore` commits JSONL to an orphan `bench-history` branch via git plumbing (versioned, pushable, shares no ancestry with `main`, never materializes in `main`'s tree). `LocalDirStore` for laptop/test use. Keyed by `(machine_id, git_commit, timestamp_unix)`; `machine_id` is a stable hash of hostname + CPU model.
- **Comparator** (`compare`): baseline = previous run on the same machine (override `--baseline <commit>`). Three regimes:
  - *distributional* (samples, n>=8, spread): two-sided Mann-Whitney U (normal approx + tie correction) **and** a 2% effect-size threshold;
  - *thresholded* (sampled but too few / zero spread, e.g. RSS): threshold only;
  - *deterministic* (no samples, e.g. alloc counts): exact compare.
  Exits non-zero on regression (CI gate); `--gate deterministic` gates only reproducible metrics.
- **Reporter** (`report`): criterion/tango-style table (value, baseline, %delta, significance marker, verdict) and `--output-json`. `viewer` is a documented stub.

## Nix integration

- `packages.indexbench` (the CLI), `apps.bench` (`nix run .#bench` → self-demo perf job), `devShells.bench` (hyperfine, valgrind, samply, jemalloc; tango is already a workspace dep), and `ix.mkBenchSuite` for consumers to declare a suite once and get both a perf-job app and a reproducible alloc-count flake check.
- Key distinction baked in: **deterministic alloc-count metrics are reproducible → `nix flake check`** (`indexbench-self-demo-alloc`); **timing/RSS are non-reproducible in the sandbox → `apps.bench` perf job**, never a flake check.

## Consumer usage

```sh
nix run index#bench                                   # self-demo perf job
nix run index#indexbench -- run --cmd "my-tool" --runs 10
nix run index#indexbench -- run --cmd "my-tool" --gate deterministic   # flake-check mode
nix run index#indexbench -- history --suite s --bench b
nix develop index#bench                               # profiling shell
```
```nix
mySuite = ix.mkBenchSuite pkgs {
  name = "search";
  indexbench = repoPackages.indexbench;
  macros = [ { name = "index-1k"; command = "file-search index ./corpus"; } ];
  allocCheck.bench = lib.getExe myAllocBenchBinary;   # optional reproducible gate
};
# mySuite.app -> apps.bench / perf job;  mySuite.check -> checks
```

## Tests & validation

- 36 unit tests (comparator significant/noise/deterministic/threshold cases, custom-metric parser, schema round-trip, machine-id stability, stores) + 2 integration tests proving the full loop: run → record to history → second run compares to previous → report.
- Validated through Nix (authoritative): `nix build .#indexbench`, the per-unit clippy / cargoAudit / cargoMachete / unusedCrateDependencies checks, every per-test-case check, and the deterministic `indexbench-self-demo-alloc` flake check all pass. `nix run .#lint` clean. Full `checks` set evaluates.

Self-demo loop output (run 1 records baselines; run 2 compares):
```
self-demo/true
  metric                candidate       baseline     delta  sig  verdict
  ----------------------------------------------------------------------
  wall_clock           321669.000     308138.500    +4.39%       unchanged
  max_rss             4370432.000    4255744.000    +2.69%  ~    unchanged
```

Please review — this is a new framework intended as the org's durable bench harness; not auto-merging.

> [!NOTE]
> ### Add `indexbench` continuous benchmarking framework with CLI, history store, and statistical comparison
> - Introduces the `indexbench` crate ([packages/indexbench/](https://github.com/indexable-inc/index/pull/497/files#diff-80683f6a1507fb750b6f513893252fd65d03ccf1d98ee43f9b1d06c3f8555af4)) providing a metric-centric benchmarking framework for both in-process micro benchmarks (wall clock, allocations) and external macro benchmarks (wall clock, RSS, custom `@bench` metrics).
> - Implements two history backends: `LocalDirStore` (JSONL file) and `GitBranchStore` (orphan git branch, CAS append to avoid concurrent clobber).
> - Adds a three-regime statistical comparator: Distributional (Mann-Whitney U with tie correction), Thresholded (±2% default), and Deterministic (exact scalar comparison), producing structured `Verdict` and `Comparison` results.
> - Provides a CLI (`indexbench run` / `assert` / `history`) that executes suites, records runs, compares against a baseline commit, and exits non-zero on gated regressions.
> - Exposes a `mkBenchSuite` Nix helper ([lib/bench.nix](https://github.com/indexable-inc/index/pull/497/files#diff-e9b231407434ada59394cb0ea50ac0ef2c0d4899d91081459fe99036bd036487)) that produces a runnable perf app and an optional deterministic alloc-check derivation; wires a `self-demo` check and `bench` app into the flake.
> - Risk: `GitBranchStore` append fails (rather than merges) on concurrent writes to the same ref, which may require retry logic in multi-job CI pipelines.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 814644b.</sup>
> 
>


andrewgazelka:
Pushed `814644b` fixing the AI-review findings (all P1/P2) plus a code-review pass.

- **store** (P1): the default git store staged its blob with an invalid `update-index --cacheinfo` form, so every append failed — replaced with `git mktree`, added compare-and-swap on the branch ref, and a round-trip test (the path had none).
- **macro_harness** (P1): drain stdout/stderr concurrently so a chatty child can't deadlock before `wait4`.
- **macro_harness** (P2): `ru_maxrss` is bytes on macOS / KiB on Linux — cfg-gate the `*1024` (was inflating darwin 1024x).
- **cli** (P2): `--cmd-name` is now a `Vec` paired with each `--cmd`; baseline is read before append (a run is never its own baseline; fixes pinned `--baseline`); first-run `--output-json` emits metrics instead of `[]`; default runs 5 → 10 with a const-assert that it clears `MIN_SAMPLES`.
- **alloc gate redesign** (P1): the two-run self-comparison could never catch a regression (both runs measure the same binary). New `indexbench assert --max metric=value` history-less budget gate; `mkBenchSuite` `allocCheck = { bench; budgets; }`. The demo now makes an exact 64 allocations and fails loudly if the allocator shim is absent.

Validated on x86_64-linux + aarch64-darwin: clippy/machete/unused-deps clean, 38 unit + 2 integration tests pass, git mktree round-trip via the CLI, assert pass/fail/missing/usage paths, a 5000-line stderr burst not deadlocking, and the demo emitting `value=64`.

Made with Claude Code (Opus 4.8).

PR #591, opened 2026-06-02
feat(dashboard): unified MIME-bundle rail for rich exec outputs (sortable DataFrames + plots)

andrewgazelka:
Exec panes had two ad-hoc rich-output rails: `images` (model-only) and, as first drafted here, `html` (dashboard-only). A third rich type would need yet another hardcoded field through worker → wire → hub → frontend. This collapses them into **one extensible `outputs` rail** modeled on Jupyter's display-data protocol.

## What you get
- **DataFrames render as a sortable HTML grid** for the operator (click-to-sort header, sticky header, dtype labels, numeric right-align, light+dark), built from the frame's own rows so it ignores any `pl.Config` cap.
- **Plots/images now render on the dashboard too** — matplotlib/PIL images were previously sent to the model but never shown to the operator.
- A single cell that `display()`s a frame *and* draws a plot produces two ordered bundles, rendered as table + image (screenshot in thread).

## Design
`outputs` is an ordered list of MIME bundles (`mime -> data`), one per displayed object / eval result / open figure. The worker builds each from the rich-display protocol it already used for images: `text/html` (df grid) + `text/plain` (repr), `image/*` base64, or whatever `_repr_mimebundle_`/`_repr_*_` offers. The frontend has a small renderer registry (`DisplayOutputs`): `text/html` → sandboxed `allow-scripts` iframe, `image/*` → img, text → pre. Adding a new rich type is frontend-only.

**Audience split:** `worker_response_content` forwards only `image/*` + the captured text to the model; rich tables are operator-only, never context. To keep the agent's captured text compact, the session applies a one-time compact `pl.Config` (20 rows/cols) the first time polars is used — and only if the user hasn't set those knobs, so their config always wins.

## Validation
- `cargo test` dashboard-core / ix-mcp / dashboard — outputs bundle round-trips wire + Loro doc
- `svelte-check` clean
- Real browser: numeric sort verified on header click; table + matplotlib plot rendered together from a real worker run (light+dark)
- `nix run .#lint` green

Linux-only clippy runs in CI.

Made with AI (Claude Opus 4.8).

> [!NOTE]
> ### Add rich MIME-bundle display rail for exec outputs with sortable DataFrame tables
> - Adds a `DisplayOutputs` Svelte component that renders rich MIME bundles (HTML, images, plain text) in a sandboxed iframe or as image/text elements, shown in exec panes and feed views.
> - Extends `ExecView` in [pane.rs](https://github.com/indexable-inc/index/pull/591/files#diff-e0dbc942625cc3362f2a451d329e414074961ccff71556697d322c34d004e3e3) with an `outputs` field (`Vec<BTreeMap<String, String>>`) carrying ordered MIME bundles, serialized to JSON and published via [hub.rs](https://github.com/indexable-inc/index/pull/591/files#diff-e4fc3a57d5c551ce7b690a0c9614f160c7b44ecd9ec9240007573f4e13501219).
> - Rewrites [python_worker.py](https://github.com/indexable-inc/index/pull/591/files#diff-7b8e5b0bdf4c09edaf2161e0febc9542d3e5e3b12a8d6f3f3525437b7e0b1710) display collection: builds per-object MIME bundles with HTML, image, and plain-text representations; generates self-contained sortable HTML table documents for polars/pandas DataFrames (capped at `MAX_HTML_ROWS`).
> - Replaces the previous flat `images` array in worker responses with an `outputs` bundle list; MCP tool results now extract image blocks from these bundles instead of a separate `images` field.
> - Behavioral Change: `worker_response_content` in [main.rs](https://github.com/indexable-inc/index/pull/591/files#diff-2dcd58b645a00e14633156b2ff05a7c89ce38e7844a856b655d1dd28aee46777) no longer reads `response.images`; workers must emit `outputs` bundles or images will not appear in tool results.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized cf6b0fb.</sup>
> 
>


andrewgazelka:
Holding this PR rather than force-rebasing it. Three blockers, in order of severity:

1. **Its backend was removed.** This PR rewrites the Rust MCP exec path (`packages/mcp/src/python_worker.py`, `exec_board.rs`, `main.rs`) to emit a MIME-bundle rail. On main those files no longer exist: the MCP was rewritten notebook-first (#640), so `packages/mcp/src` is now just `screen/` + `vmkit/` Python helpers. The mcp-side of this PR targets deleted files and would need a ground-up re-port against the new notebook-first MCP, assuming the sortable-DataFrame feature is still wanted there.
2. **Non-linear history.** The branch carries a `Merge branch 'main'` commit; the main ruleset requires linear history, so it can't merge without a rebase anyway.
3. **Open security finding (needs a decision).** The AI review flagged that `DisplayOutputs.svelte` renders every `text/html` bundle in an iframe with `sandbox="allow-scripts"`, so arbitrary `_repr_html_`/`_repr_mimebundle_` output (not just the trusted table) can run JS and make network requests. Interactive itables legitimately need JS, and the dashboard is operator-facing in a single-tenant sandbox, so this may be acceptable by design, but it's a call to make explicitly before shipping.

Recommendation: decide whether the feature still belongs in the notebook-first MCP. If yes, re-implement it there (the `ExecBody.svelte` rendering side may still apply) with the HTML payload bounded (cap columns / per-cell length / total bytes, not just `MAX_HTML_ROWS`) and a deliberate stance on the iframe sandbox. If no, close this.

andrewgazelka:
Superseded by the notebook-first MCP rewrite (#640) and itables (#669), which re-implement this PR's entire feature set a cleaner way:

- **Sortable DataFrames** — `packages/mcp/ix_notebook_mcp/ipython/00-ix-itables.py` enables itables, so every pandas/polars DataFrame/Series auto-renders as an interactive sortable table at the IPython formatter level (with a `text/plain` fallback kept for the agent path and non-JS viewers). That replaces this PR's bespoke DataFrame->HTML table in the Rust `python_worker.py`.
- **Plots** — `packages/mcp/ix_notebook_mcp/outputs.py` returns figures as image blocks (`image/png`/`jpeg`).
- **The unified MIME rail** — handled now by nbformat outputs plus the dashboard's renderer registry (`packages/dashboard-core/site/src/lib/renderers.ts`: `html -> HtmlBody`, plus `DataBody`/`ExecBody`).

The Rust MCP files this PR edits (`python_worker.py`, `exec_board.rs`, `main.rs`) no longer exist on main, and the feature they delivered is already present, so there's nothing left to re-port. Closing.

The two review findings are moot here but worth carrying into the new code if not already handled: (1) bound the itables HTML payload (columns / cell length / total bytes, not just rows), and (2) the `text/html` iframe `sandbox="allow-scripts"` posture (needed for itables JS; acceptable for the single-tenant operator dashboard, but an explicit call).

PR #654, opened 2026-06-03
feat(search): default to mxbai-rerank-v3-listwise reranker

andrewgazelka:
Wire Mixedbread's new listwise reranker (`mxbai-rerank-v3-listwise`) through the whole search stack and make it the default.

- `mixedbread::Rerank` replaces the bare `rerank: bool`. It serializes as the API's `boolean | object` field: a toggle keeps the legacy bare-bool wire body, a model selection sends `{ "model": ... }`. Wire shape locked by a new serde test.
- CLI (`search`/`mgrep`), the MCP/`search-py` binding, and `polars-mixedbread` now default to the listwise model. `--no-rerank` still disables; `--reranker <model>` / `reranker=` pins another model.
- `search-eval` gains `--reranker` so the listwise default can be measured against a baseline.

The listwise model reads the candidate set as a whole instead of scoring each chunk independently, which lifts ranking quality over the prior pointwise default (Mixedbread reports gains on every benchmark they tested).

Validated: `nix build .#search .#polars-mixedbread .#file-search .#indexer` green on aarch64-darwin (compile gate covers mixedbread + search-core + the polars binding). Linux rust checks (including the Linux-only `search-py` wheel) run in CI.

Made with Claude (Opus 4.8).

> [!NOTE]
> ### Default reranking to `mxbai-rerank-v3-listwise` across search CLI and Python bindings
> - Introduces a `Rerank` enum in [packages/mixedbread/src/lib.rs](https://github.com/indexable-inc/index/pull/654/files#diff-1ccc2330cdefa0e3534bffd0e0b7af64c448d14a8f6715cc4f69c594032e607e) with variants for off, server default, and named model; serializes to boolean or `{model: ...}` to match the API contract.
> - Adds `DEFAULT_RERANK_MODEL` constant (`mixedbread-ai/mxbai-rerank-v3-listwise`) used as the default when reranking is enabled without an explicit model.
> - Updates the search CLI (`--reranker` flag, defaults to `DEFAULT_RERANK_MODEL`) and Python `semantic()`/`scan_mixedbread()` bindings to accept an optional reranker model name.
> - Behavioral Change: enabling reranking without specifying a model previously used the server default; it now pins `mxbai-rerank-v3-listwise` explicitly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 76c9520.</sup>
> 
>

### DIFF
--- a/packages/mixedbread/src/lib.rs
+++ b/packages/mixedbread/src/lib.rs
@@ -122,14 +122,66 @@ pub enum Error {
 /// Result alias defaulting to this crate's [`Error`].
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
+/// Default second-stage reranking model: Mixedbread's listwise reranker.
+///
+/// It reads the candidate set as a whole and lifts ranking quality across every
+/// benchmark relative to the prior pointwise default.
+/// See <https://www.mixedbread.com/blog/mxbai-rerank-v3-listwise>.
+pub const DEFAULT_RERANK_MODEL: &str = "mixedbread-ai/mxbai-rerank-v3-listwise";
+
+/// Second-stage reranker selection.
+///
+/// Serialized as the API's `rerank` field, which is `boolean | object`:
+/// [`Rerank::Toggle`] serializes to a bare bool (so the legacy "just turn it
+/// on/off" wire body is byte-for-byte unchanged), while [`Rerank::Model`]
+/// serializes to `{ "model": "..." }` to pin a specific reranking model.
+#[derive(Debug, Clone, serde::Serialize)]
+#[serde(untagged)]
+pub enum Rerank {
+    /// Toggle the server's default reranker: `false` disables reranking,
+    /// `true` applies the API-chosen default model.
+    Toggle(bool),
+    /// Apply a named reranking model, e.g. [`DEFAULT_RERANK_MODEL`].
+    Model {
+        /// Reranking model name forwarded to the API.
+        model: String,
+    },
+}
+
+impl Rerank {
+    /// Reranking disabled (wire `false`).
+    #[must_use]
+    pub const fn off() -> Self {
+        Self::Toggle(false)
+    }
+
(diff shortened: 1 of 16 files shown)


andrewgazelka:
Holding this after rebasing onto current main. The rebase exposed a real collision that needs careful, compile-validated reconstruction (not a blind re-push):

**What happened.** Main refactored `mixedbread::SearchOptions` (added `score_threshold` / `return_metadata`, kept `rerank: bool`). This PR replaces `rerank: bool` with a new `Rerank` enum (`Toggle(bool) | Model { model }`, untagged so the wire body stays `bool | object`) plus `DEFAULT_RERANK_MODEL`. During the rebase the `mixedbread/src/lib.rs` (and `polars-mixedbread/src/lib.rs`) conflict got resolved to main's side, **dropping the `Rerank` type entirely**, so `search-core/src/lib.rs` fails to build: `error[E0432]: unresolved imports mixedbread::DEFAULT_RERANK_MODEL, mixedbread::Rerank`.

**What finishing it needs** (the API-trailing-param fix from the AI review is already applied and reviewed-clean):
1. Re-add `Rerank` + `DEFAULT_RERANK_MODEL` + impl to `mixedbread/src/lib.rs` (verbatim from commit 53cf31c4), change `SearchOptions.rerank: bool` → `Rerank`, drop `Copy` from that derive, and **keep** main's `score_threshold`/`return_metadata` fields.
2. Re-verify the auto-merged `search-core/{adapter,backend,search}.rs`, `polars-mixedbread/src/lib.rs`, and the 5 `SearchOptions { ... }` construction sites set both `rerank: Rerank` and main's new fields.
3. Validate on x86_64-linux (the shipped `rerank_serializes_as_bool_or_object` test guards the wire format) — local darwin can't build the `.#checks.x86_64-linux.*` units.

I didn't rush this blind because it's wire-format-sensitive and main is churning, which makes CI-only iteration unreliable. Original pre-rebase commit for reference: 39c305a2.

PR #693, opened 2026-06-04
refactor(rust): build pipeline cleanup and nix lint idioms (stage 1 of N)

anna328p:
## Summary

Stage 1 of the Rust build-infrastructure refactor. It restructures the `lib/rust/` build pipeline for clearer ownership and fewer hand-rolled abstractions, introduces a batch of ast-grep Nix rules that codify the idioms it adopts, and applies those idioms across the repo. This is a structural change with no intended behavior difference: the affected packages, checks, and per-crate policy gates eval to identical derivations.

## Rust build pipeline (`lib/rust/`)

`cargo-unit.nix` and `build.nix` are the public surface for the per-unit Rust build pipeline: workspace unit graphs, vendoring, policy checks, and package builds. Both had grown freeform option resolvers that silently swallowed misspelled fields, normalized the same args repeatedly across the call chain, and scattered single-use helpers away from their call sites.

- A single `normalizeArgs` resolver, shared across `build.nix` and `cargo-unit.nix`, applies policy and vendor defaults once and threads the result onward so nothing downstream re-normalizes. It enumerates policy fields, so a typo'd key fails loudly instead of being silently dropped.
- `clippyLintFlagsFromManifest` moves next to `clippyLintArgs`; single-use helpers move to their one caller; `cargoLockFile` collapses to `cargoLock.lockFile or cargoLock`; the dead `useNextest` fallback is removed; builtins are hoisted via `inherit (builtins)`.
- `rustcArgsForPolicyForPlatform` no longer tolerates a `null` platform sentinel. Its callers in `build.nix` pass a real triple, and `cargo-unit`'s `extraRustcArgsForPlatform` resolves the renderer's per-unit `null` (host units: build scripts and proc-macros) to `hostPlatform.config` at the seam.
- `pname` is required (or derived from `name`); a missing name is an eval-time error rather than a silent default.

## ast-grep Nix rules

New rules: `no-chained-attrset-update` (`A // { ... } // { ... }` becomes one merged update), `prefer-sorton-over-keyed-sort` (`sort (a: b: f a < f b) xs` becomes `lib.sortOn f xs`, evaluating the key once per element), plus `prefer-genlist-over-map-range`, `prefer-or-default-over-has-attr-guard`, and `prefer-imap0-over-genlist-identity` from earlier in the branch.

Tightened rules: `prefer-genattrs-listtoattrs` now also catches the keyed-by-element `genAttrs` case, and `no-deprecated-iflist-empty` adds the swapped `if cond then [ ] else y` orientation. `fleet.nix`'s cycle detection is exempted from the latter with a documented reason: its non-empty branch returns the first offending cycle, not list elements to splice.

The earlier rule consolidation folded the legacy hex-`sha256` checks into one `prefer-sri-hash` and dropped a dead `pkgs.lib` branch from `no-bare-assert` that `no-pkgs-lib` already made unreachable.

## Repo-wide idiom fixes

The rules drive mechanical conversions across the tree: `listToAttrs (map …)` becomes `genAttrs'`/`genAttrs` in agent-context, discovery, overlay, per-system, minecraft, and registry; a `lib.sort` keyed comparator becomes `lib.sortOn` in the minecraft-blocks schema; and `if x == null then [ ] else …` becomes `lib.optionals` in discovery.

## Verification

`nix run .#lint` passes all five stages (the pre-commit hook runs it on each commit). drvPath equality holds across the build graph: `dag-runner`, `ix-vt`, both `-package` checks, `agents-md-package`, `unusedCrateDependencies`, and `cargo-unit-prebuilt-library` eval byte-identical to the pre-refactor baseline. The four policy-check attrs (`rust-cargoAudit`, `rust-agents-md-clippy`, `llm-clippy`, `nix-cargo-unit`) differ only by one removed trailing blank line in the policy-check scripts. The per-crate policy gates survive: `passthru.tests` still exposes `clippy` and `unusedCrateDependencies` alongside the per-`#[test]` derivations.

> [!NOTE]
> ### Clean up Nix build pipeline and add ast-grep lint rules for idiomatic Nix patterns
> - Refactors [lib/rust/cargo-unit.nix](https://github.com/indexable-inc/index/pull/693/files#diff-82054df1a7d1648a5be88265641a7ae5d3b9e4ec4cdea72c518359ea4ae3f804) and [lib/rust/build.nix](https://github.com/indexable-inc/index/pull/693/files#diff-8ab3484d4c25eaf4ab0beac36848ba87229282d10d3524b969f90a880b330dab): inlines formerly separate `generateUnitGraph`/`generateUnitsNix` helpers into `buildWorkspace`, consolidates vendoring and policy resolution into a new `normalizeArgs` function, and removes previously exported `auditCargoLock`, `generateUnitGraph`, `generateUnitsNix`, `vendorDir`, `vendorSources`, `resolvePolicy`, `resolveVendorSources`, and `resolveVendorDir`.
> - Adds a new `clippyLintFlagsFromManifest` helper in [lib/rust/build.nix](https://github.com/indexable-inc/index/pull/693/files#diff-8ab3484d4c25eaf4ab0beac36848ba87229282d10d3524b969f90a880b330dab) that derives per-unit clippy flags from a Cargo.toml without running `cargo clippy`.
> - Adds ~15 new ast-grep rules under [ast-grep/nix/rules/](https://github.com/indexable-inc/index/pull/693/files#diff-c1086742a1ddc1ff172ba30232da0fd4464e1a7714b01bee2538f94aa1a8126b) enforcing idiomatic Nix patterns: preferring `lib.genAttrs`/`lib.sortOn`/`lib.imap0`/`lib.importJSON`, SRI hashes over legacy `sha256` fields, `or` defaults over `hasAttr` guards, and flagging chained attrset updates and empty-branch conditionals.
> - Applies the new lint rules across the codebase, replacing `listToAttrs (map ...)` with `lib.genAttrs'`, `lib.sort` with `lib.sortOn`, and `builtins.*` with imported or prefixed equivalents in multiple files.
> - Adds a `devShells.default` in [lib/per-system.nix](https://github.com/indexable-inc/index/pull/693/files#diff-9879a1f03396eb758ff538e44a00fb409b99bc0bbfac4656755b05716d036683) providing `ast-grep` and `nixfmt`.
> - Risk: callers of `lib/rust/cargo-unit.nix` or `lib/rust/build.nix` that referenced removed exports (`auditCargoLock`, `resolveVendorDir`, etc.) will fail at evaluation time. `buildBinary`/`selectBinaryWithTests` now throw on missing binaries instead of falling back to `workspace.default`; caller-provided `useNextest` is no longer honored.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0769aac.</sup>
> 
>


anna328p:
to elaborate: i have spent the past two days scanning antipatterns in the codebase and setting up structure - pr #687 distills pretty much all of it. it sets up for a major refactor that i am now letting fall into place; you should see a bunch of new commits come in, applying the new rules to the current state of the repo, and at the same time there's a huge class of design errors that will no longer exist in newly generated code (if the appropriate skill gets invoked)

anna328p:
going to rebase locally instead of keeping those merge commits

PR #750, opened 2026-06-05
Nest per-#[test] checks so nix-eval-jobs shards enumeration instead of ballooning one worker

harivansh-afk:
> TLDR;
>
> the rust checks were a flat per-#\[test\] set, so nix-eval-jobs had to force every crate's manifest in one worker, ballooning it to 30-42 GiB until earlyoom killed it on the shared runner.
>
> The PR keeps the flat checks output (so nix flake check still works) but adds a sharded ciChecks that nests each package under recurseForDerivations, so the evaluator forces one crate at a time in restartable workers instead of all at once.

## Problem

`blast-radius` (and the `check` gate) evaluate `.#checks.x86_64-linux` with `nix-eval-jobs`. The rust checks are flattened into the top-level `checks` set by name, so to list the attr names the single worker assigned the root attrpath must force every crate's per-`#[test]` manifest IFD at once (the manifest builds each test binary to enumerate its `#[test]` cases). That one worker balloons and cannot restart, because the memory cap is only checked between jobs.

Observed live on the shared CI host (`vin-compute-1`): one `nix-eval-jobs` worker growing 14 -> 27 -> 42 GiB on a single job while sibling workers stayed \~4 GiB. When several blast-radius jobs run at once the host drops below 5 percent free and `earlyoom` SIGKILLs the evaluators (13 such kills in one incident window, all `nix-eval-jobs`), so the check fails non-deterministically. The cgroup never OOMs (`oom_kill 0`); earlyoom fires host-wide first.

This is the per-crate split the `checks` block already documents, but `mergeAttrsList` over renamed per-case keys flattened it back, so enumeration stayed monolithic.

## Fix

Nest each package's tests under one attr carrying `recurseForDerivations` instead of flattening them into the top-level set. Listing `checks` now returns the cheap per-package names (no manifest forcing), and each crate's manifest IFD is forced inside its own worker job, which restarts at the memory cap between packages. Peak per worker drops to roughly the per-crate floor (\~4 GiB), and aggregate becomes `workers x cap`, which is what makes `--max-memory-size` and worker restart effective and keeps earlyoom from seeing a tens-of-GiB target.

The leaf derivations are byte-identical, so the per-case count and `--timings` keys are unchanged; check names gain a per-package path segment (`rust-foo.bar-tests-baz` instead of `rust-foo-bar-tests-baz`). The gate (nix-fast-build) and blast-radius both wrap `nix-eval-jobs`, which recurses `recurseForDerivations`, so both see the same names consistently.

A subtle point captured in the code comment: the nested value must stay a thunk. Filtering empty groups (for example `tests != {}`) would force every package's tests during enumeration and reintroduce the balloon, so empty groups are left in and nix-eval-jobs simply recurses them to nothing.

## Validation

* Evaluation succeeds locally: lint ran through eval and started building the cargo-units IFD. It then failed only because this developer's nix daemon lacks the `ca-derivations` experimental feature (non-trusted user, cannot override), which is why the local pre-commit lint hook was skipped. CI runs the identical lint and the `check` gate with `ca-derivations` and builders.
* Laziness of the nesting is the load-bearing property and follows from Nix semantics: `attrNames` forces attrset keys, not values.
* Final memory confirmation is the gate run on this PR, plus the same one-line pattern applied to the `ix` repo's `#checks` assembly (the catalog seen ballooning on prod is the `ix` repo, a separate consumer of this lib; that mirror is the prod-landing step).

Part of [ENG-2201](https://linear.app/indexable/issue/ENG-2201/blast-radius-eval-balloons-one-nix-eval-jobs-worker-and-gets-earlyoom)

> \[!NOTE\]
>
> ### Nest per-`#[test]` checks under `recurseForDerivations` groups in `ciChecks` so nix-eval-jobs shards by package
>
> * Adds a new `ciChecks` flake output where Rust checks are grouped per-package using `recurseForDerivations`, so `nix-eval-jobs` evaluates each package shard independently rather than enumerating all leaf derivations in a single worker.
> * Introduces `rustPackageTestSets` with `flat` and `sharded` variants; `checks` continues to use the flat keying (one derivation per test), while `ciChecks` uses the sharded keying.
> * Adds a `catalogFor` helper in [flake.nix](<https://github.com/indexable-inc/index/pull/750/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0>) that centralises composition of explicit, Rust, and image checks with collision-safety assertions.
> * Updates `nix run .#check`, the blast-radius diff, and CI workflows to target `.#ciChecks.x86_64-linux` instead of `.#checks.x86_64-linux`.
> * Behavioral Change: `eval_checks` in [packages/blast-radius/src/nix.rs](<https://github.com/indexable-inc/index/pull/750/files#diff-14eaffb995110afd87ea035ed088cee05190e63275835359b273ee0fce584a12>) now calls `nix-eval-jobs` against `#ciChecks.x86_64-linux`; the leaf derivations are the same but the evaluation topology changes.
>
> [Macroscope](<https://app.macroscope.com>) summarized b455383.

> [!NOTE]
> ### Nest per-`#[test]` checks under package attrs so nix-eval-jobs shards enumeration across workers
> - Adds a `ciChecks` flake output alongside `checks`, where Rust tests are grouped per-package (`recurseForDerivations`) instead of flattened, letting nix-eval-jobs distribute evaluation across workers rather than serializing it in one.
> - Introduces `rustPackageTestSets` with `flat` and `sharded` views of the same leaf derivations; `checks` uses `flat`, `ciChecks` uses `sharded`.
> - Updates `nix run .#check` and blast-radius to target `.#ciChecks.x86_64-linux` instead of `.#checks.x86_64-linux`.
> - blast-radius now probes both revisions via `catalog_attr` and picks `ciChecks` when both have it, ensuring base and head use the same attribute keying for diffs.
> - Adds `normalize_attr` in [nix.rs](https://github.com/indexable-inc/index/pull/750/files#diff-14eaffb995110afd87ea035ed088cee05190e63275835359b273ee0fce584a12) and [timings.rs](https://github.com/indexable-inc/index/pull/750/files#diff-d01a843a7f3d06afa53965b354dadf9a47a5c27068a4de3ba5511fd0a69ae8d0) to strip quotes from nix-eval-jobs' nested attribute path encoding before diffing or indexing.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6747938.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -39,7 +39,7 @@ jobs:
       github.event.pull_request.user.login != 'dependabot[bot]'
     # Same self-hosted dispatcher as check.yml: the `ix-ci-run-*` label is
     # claimed by the org dispatcher, which mints an ephemeral runner with a warm
-    # /nix/store. Evaluating .#checks.x86_64-linux forces an x86_64-linux IFD
+    # /nix/store. Evaluating .#ciChecks.x86_64-linux forces an x86_64-linux IFD
     # build (lib/rust/cargo-unit.nix), so this must run on the Linux runner.
     runs-on: ["${{ format('ix-ci-run-{0}-{1}-blast-radius', github.run_id, github.run_attempt) }}"]
     timeout-minutes: 30
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -77,7 +77,7 @@ jobs:
 
       # Run the full gate through the repo-owned `check` app (lib/per-system.nix),
       # so CI and a local `nix run .#check` execute the same two commands from one
-      # definition: nix-fast-build over `.#checks.x86_64-linux`, then nix-eval-jobs
+      # definition: nix-fast-build over `.#ciChecks.x86_64-linux`, then nix-eval-jobs
       # over `.#packages.x86_64-linux` with the JSON error-line gate. The
       # command-specific rationale (tool choice, the worker/memory tuning, why the
       # eval cache is off, the error-line gate) and the pinned tool revisions live
--- a/flake.nix
+++ b/flake.nix
@@ -208,6 +208,12 @@
       overlays.default = ix.overlay;
       packages = collect "packages";
       checks = collect "checks";
+      # Sharded keying of the same check derivations for the memory-bounded CI
+      # evaluator (the `.#check` gate and blast-radius); see lib/per-system.nix
+      # (ENG-2201). Kept separate from `checks` because its per-package
+      # `recurseForDerivations` groups are not derivations, which the flake
+      # `checks` schema requires.
+      ciChecks = collect "ciChecks";
       formatter = collect "formatter";
       apps = collect "apps";
       devShells = collect "devShells";
--- a/lib/per-system.nix
+++ b/lib/per-system.nix
@@ -108,7 +108,7 @@ let
   # nix is already present); pinning a client nix here could mismatch the host
   # Nix 2.34.x daemon.
   #
-  # Step 1 (nix-fast-build) builds every `checks.x86_64-linux` derivation: it
+  # Step 1 (nix-fast-build) builds every `ciChecks.x86_64-linux` derivation: it
   # evaluates with nix-eval-jobs (parallel) and streams each drv into a build
   # pool as it resolves. --skip-cached drops paths already in a substituter (a
   # warm run does almost no work), --no-nom keeps plain logs, --no-link leaves no
@@ -139,7 +139,7 @@ let
   # host Nix 2.34.x.
   check = ix.writeNushellApplication pkgs {
     name = "check";
-    meta.description = "Run the full CI gate: build .#checks.x86_64-linux and eval-validate .#packages.x86_64-linux";
+    meta.description = "Run the full CI gate: build .#ciChecks.x86_64-linux and eval-validate .#packages.x86_64-linux";
     text = ''
       const fast_build = "github:Mic92/nix-fast-build/7f185e0ec37b65b4730f892e0de9a831b0610f3a"
       const eval_jobs = "github:nix-community/nix-eval-jobs/65ebf5b7cd453a27af09cf02b1fc57b3568cc4b7"
@@ -147,7 +147,7 @@ let
       def main [] {
         # ca-derivations: the rust workspace units default to
         # `contentAddressed = true` (lib/rust/cargo-unit.nix), so evaluating
-        # `.#checks.x86_64-linux` resolves floating content-addressed drvs. The
+        # `.#ciChecks.x86_64-linux` resolves floating content-addressed drvs. The
         # evaluator (nix-eval-jobs, which nix-fast-build wraps) needs the
         # `ca-derivations` experimental feature, or it aborts with
         # "experimental Nix feature 'ca-derivations' is disabled". The flake's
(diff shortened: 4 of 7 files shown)


harivansh-afk:
Addressed the P1 (nested checks break the flake schema) in b455383.

Confirmed the issue locally: a non-derivation under `checks.x86_64-linux.<name>` makes `nix flake check` error `flake attribute ... is not a derivation`. Fix: `checks` stays flat and schema-valid; the sharded keying (one `recurseForDerivations` group per package) moved to a separate `ciChecks` output, which `nix flake check` validates-skips (`unknown flake output`). The `.#check` gate and blast-radius now evaluate `.#ciChecks.x86_64-linux`. Both keyings come from one `catalogFor` over the same leaf derivations, so the per-#[test] contract (count, names, `--timings` keys) is unchanged, and the collision guard runs per keying so evaluating `ciChecks` never forces the flat per-#[test] spine.

Note for the ix mirror: blast-radius now targets `.#ciChecks`, so the ix repo must add a `ciChecks` output before bumping this flake input.

harivansh-afk:
Addressed the second P1 (base revs predating `ciChecks`) in 61a94dd.

`eval_checks` now probes the flake output set cheaply (`nix eval ...#ciChecks --apply builtins.isAttrs`, which forces only the `{ <system> = ...; }` spine, not the catalog) and targets `ciChecks` when present, else falls back to the flat `checks`. So head (which has `ciChecks`) evaluates the sharded catalog, and an older merge base still evaluates via `checks`. Both outputs share the same leaf derivations, so the diff is identical. Flagged in-code as a migration shim to remove once no evaluated base predates `ciChecks`. `cargo check` passes locally.

harivansh-afk:
Addressed the mixed-keying P1 in a892df8.

The catalog attribute is now resolved once (`nix::catalog_attr` over both base and head) and used for both evals: `ciChecks` when both revisions expose it, else flat `checks` for both. Base and head are therefore always keyed identically, so unchanged derivations diff as unchanged rather than removed+added. The flat fallback only engages while a PR's merge base predates `ciChecks`; once this lands on main, both sides use `ciChecks` and it is a no-op. `cargo check` passes.

harivansh-afk:
Addressed the quoted-leaf-segment P1 in 6747938.

Root cause: nix-eval-jobs joins a nested attr path with `.` and wraps any single segment that contains a `.` in double quotes. A sharded `ciChecks` doctest leaf carries a file path in its case name (`src/lib.rs - (line 12)`), so it surfaces as `rust-foo."doctest-...src/lib.rs - (line 12)"`: bare package segment, quoted leaf. The old `trim_matches('"')` only stripped the ends, so the embedded `"` survived. nix-fast-build copies that joined string verbatim into its `--timings` records (it ignores the `attrPath` array), and the trusted workflow schema (`blast-radius.yml` safename regex `^[A-Za-z0-9_./ +():-]+$`) rejects `"`, so one such key in a base timings artifact could fail the comment job even when that check did not change.

Fix: added `nix::normalize_attr`, which drops every `"` and splits on dots outside quotes, then rejoins with `.`. Used in both the eval parser (`partition_eval_rows`) and the timings parser (`timings::parse`), so the bare dot path (`rust-foo.doctest-...src/lib.rs - (line 12)`) flows identically through the diff key, the report, and the schema. The regex allows dots, slashes, spaces, and parens, so the unquoted name validates. A flat single-segment name (quoted whole because its case name holds a `.`) round-trips to the same bare name the old end-trim produced, so non-sharded keying is unchanged. Added unit tests for both parsers covering flat, sharded-quoted, and sharded-unquoted shapes; `cargo test -p blast-radius` is green (14 passed).

Issue #752, opened 2026-06-05
Sanity-check the ingestion architecture: one log, ClickHouse + Mixedbread views


andrewgazelka:
Is the one-log-many-views design the right spine for ingesting domain facts (Minecraft block placements, Claude/Codex history, shell history, journald) and fanning out to both ClickHouse and Mixedbread? Want a sanity check on the architecture before we lean on it harder.

< />< />< />mermaid  <br>flowchart LR  <br>subgraph SOURCES< />"SOURCES (producers)"< />  <br>blocks< />"Minecraft block< />nplacement (fact)"< />  <br>chats< />"Claude / Codex< />ntranscripts"< />  <br>shell< />"shell history< />n+ journald"< />  <br>code< />"code repos"< />  <br>end

subgraph TRANSPORT< />"TRANSPORT (lossy delivery)"< />  <br>kprod< />"tail + kafka-producer"< />  <br>otel< />"OTel collector< />n(filelog + journald)"< />  <br>end

subgraph LOG< />"LOG (durable, replayable truth)"< />  <br>topic< />"Kafka/Redpanda topic"< />  <br>parquet< />"Parquet/Iceberg on S3/R2"< />  <br>end

subgraph VIEWS< />"VIEWS (one per query pattern)"< />  <br>ch< />"ClickHouse< />ncolumnar / range"< />  <br>mxbai< />"Mixedbread< />nvector index"< />  <br>end

grafana< />"Grafana"< />  <br>mgrep< />"search / mgrep"< />

blocks --> kprod --> topic  <br>chats --> otel --> topic  <br>shell --> otel  <br>topic --> parquet  <br>parquet --> ch  <br>parquet -->|indexer reads log, dedups, pushes docs| mxbai  <br>code -->|documents, not events: skip the log| mxbai  <br>ch --> grafana  <br>mxbai --> mgrep  <br>< />< />< />

Current shape:

* Transport (kafka-producer tail, OTel collector) is lossy delivery, not truth.
* One durable log (Kafka/Redpanda topic materialized to Parquet/Iceberg on object storage) is the source of truth.
* Views derived from the log: ClickHouse for columnar/range/aggregate (Grafana on top), Mixedbread for semantic search (search/mgrep on top). Mixedbread is fed from the S3/Parquet log via source/parquet + sink/parquet so an outage resumes from the buffer.
[...]

PR #786, opened 2026-06-07
tools: iceberg-rust 0.9 spike for the R2 corpus log

wyattgill9:
Phase 0.4 of the #752 plan: prove the `sink-iceberg`/`source-iceberg` design against iceberg-rust 0.9.1 with running code before building the real crates (which landed as `lake-iceberg` in #785 — this spike is the instrument that shaped it).

Standalone workspace on purpose (own lockfile, like `packages/nix-cargo-unit`): the throwaway never touches the root lock, unit graph, or lint wall.

All six scenarios pass against the in-memory catalog: namespace+table creation, two append commits, full scan, snapshot-cursor incremental read, concurrent writers. Findings baked into comments and the README, each of which became a design rule in #785:

1. Per-run UUID file-name suffix is mandatory (a reused name is rejected as already-referenced).
2. The incremental cursor is a ~30-line manifest walk we own: `sequence_number` cursor, `Append`-only snapshots (`Replace` = compaction would re-deliver the table), `added_snapshot_id` filter (manifests are immutable and carried forward).
3. REST catalog in 0.9 needs an explicit `OpenDalStorageFactory` and the `"token"` prop; standard `iceberg::io::S3_*` props configure the data plane.

The R2 half (real auth, managed compaction vs the cursor, snapshot expiration) pends Cloudflare credentials; the same binary runs against R2 with `SPIKE_CATALOG=rest` per the README. The local REST fixture lives with the lake crate in #785 (`packages/lake/iceberg/fixture/`).

Refs #752.

> [!NOTE]
> ### Add iceberg-rust 0.9 spike tool for validating Iceberg table operations against R2
> Adds a standalone Rust binary at [tools/iceberg-r2-spike](https://github.com/indexable-inc/index/pull/786/files#diff-82c9b473004d357a8c1ab8570457f800298f453b393084abd54c59d15df4a7b5) that exercises iceberg-rust 0.9 against an in-memory or REST/R2 catalog to validate design assumptions before production use.
>
> - Implements six scenarios (S1–S6): namespace/table creation, sequential appends, full-scan row count, incremental snapshot cursor reads, and concurrent optimistic-commit writes with retry
> - Supports two catalog modes via `SPIKE_CATALOG` env var: in-memory with a temp file warehouse (default) or a REST catalog backed by OpenDal S3 storage for R2
> - The `added_files_since` cursor logic filters manifest entries by snapshot sequence number and `Added` status, excluding compaction (`Replace`) operations
> - Includes a [README.md](https://github.com/indexable-inc/index/pull/786/files#diff-aa785bfca29214e46ce07faf57aebeda5fc4d9704dbccc84e6cb3547a6bd85ca) with findings, run instructions, and remaining verification tasks
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3fbf3c9.</sup>
> 
>


wyattgill9:
Throwaway spike — it did its job: validated iceberg-rust 0.9.1 (REST catalog, append, commit-conflict retry, incremental scan). The corpus lake ships on that validation in #785 + the leader fold, standing on internal MinIO rather than R2 (see the internal-S3 pivot on #752). Closing without merging; nothing here is meant for main.

PR #874, opened 2026-06-08
feat(cargo-unit): opt-in cargoConfigRustflags to apply .cargo/config.toml rustflags

andrewgazelka:
## What

Add an opt-in `cargoConfigRustflags` to `cargoUnit.buildWorkspace` that applies the rustflags a normal `cargo build` reads from `.cargo/config.toml`.

## Why

cargoUnit builds one derivation per rustc unit and assembles rustc args itself (`policy` + `extraRustcArgs` + `extraRustcArgsForPlatform`); it never reads `.cargo/config.toml`. So target rustflags cargo would apply, e.g.

```toml
[target.aarch64-apple-darwin]
rustflags = ["--cfg", "reqwest_unstable", "-Zthreads=8"]
```

are silently dropped. Downstream this forces hand-copying flags into `extraRustcArgs`: the [nox](https://github.com/indexable-inc/nox) flake had to pass `--cfg reqwest_unstable` by hand (reqwest's `http3` feature hard-errors without it), duplicating what its `.cargo/config.toml` already declares.

## How

- `lib/rust/build.nix`: new `rust.rustflagsFromCargoConfig configPath platform`. Parses via `lib.importTOML`; returns rustc args for a target triple with cargo precedence (`target.<triple>.rustflags` wins outright over `build.rustflags`, no merge; array or whitespace-string). Missing config -> `[]`.
- `lib/rust/cargo-unit.nix`: new `cargoConfigRustflags ? false`. When set, appends the parsed flags into the existing per-unit `extraRustcArgsForPlatform` seam (platform-aware, exactly cargo's model). **Default off**, so every current consumer is unchanged.
- Out of scope (documented): `cfg(...)` target sections and the `[env]` table.

## Test

New zero-dependency fixture `tests/fixtures/cargo-unit-cargo-config/` whose crate has `#[cfg(not(cargo_config_ok))] compile_error!` plus a `.cargo/config.toml` setting that cfg. `cargoUnitCargoConfigWorkspace` builds it with the option on and the helper aggregate runs the binary: it compiles only because the option fed `--cfg cargo_config_ok`, so the test is behavioral, not a string assertion on argv.

## Validation (aarch64-darwin)

- Helper precedence verified by `nix eval` (target overrides build; `cfg(...)` ignored; missing file -> `[]`).
- Real build with the option **on** succeeds and the rustc invocation carries `--cfg cargo_config_ok`; with it **off** the build hits the crate's `compile_error`.
- `nixfmt` clean.

Follow-up (separate, not here): once this lands and nox bumps its `index` pin, replace nox's `extraRustcArgs = ["--cfg" "reqwest_unstable"]` with `cargoConfigRustflags = true`.

> [!NOTE]
> ### Add opt-in `cargoConfigRustflags` to pass `.cargo/config.toml` rustflags to rustc in `cargoUnit.buildWorkspace`
> - Adds a `rustflagsFromCargoConfig` helper in [build.nix](https://github.com/indexable-inc/index/pull/874/files#diff-8ab3484d4c25eaf4ab0beac36848ba87229282d10d3524b969f90a880b330dab) that reads `[build].rustflags` or `[target.<triple>].rustflags` from a Cargo config file, honoring cargo precedence (target-specific wins, no merging), and normalizes them to rustc args.
> - Adds a `cargoConfigRustflags` option in [cargo-unit.nix](https://github.com/indexable-inc/index/pull/874/files#diff-82054df1a7d1648a5be88265641a7ae5d3b9e4ec4cdea72c518359ea4ae3f804); when `true`, flags from `<workspaceRoot>/.cargo/config.toml` are appended after all other rustc args so explicit caller args still take precedence.
> - Adds a test fixture and build in [tests/default.nix](https://github.com/indexable-inc/index/pull/874/files#diff-1cc580de297308d93d82f7b72446ae4b98832a8aae3378e9e134102519a0e33a) that uses `--cfg cargo_config_ok` to verify the flags are applied at compile time.
> - Behavioral Change: when `cargoConfigRustflags = false` (the default), behavior is unchanged; opting in may introduce rustflags that conflict with policy-derived flags if the config contains unexpected entries.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e2111a8.</sup>
> 
>


andrewgazelka:
note here : Zthreads 8 is not determinitic and it shoudl be imrpoved etc

PR #878, opened 2026-06-08
fix(mcp): bound python_exec budget and keep the shell channel clean on cancel

andrewgazelka:
## What
A `python_exec` budget is how long a run holds the kernel's one shared shell channel before it backgrounds, but nothing bounded it, and a client cancel could desync that channel. Two failure modes, one root area:

1. **Oversized budget blocks everything.** `await jobs['x']` with `budget=900` held the channel for 15 minutes, so every other `python_exec` (including status checks) blocked behind it.
2. **Cancel wedges the kernel.** Cancelling an in-flight `python_exec` threw `CancelledError` straight into `execute_interactive`, abandoning a half-read multipart reply on the shell socket. The socket desynced and every later call hung. This is the "I did mcp cancel and now nothing runs" wedge.

## Changes
- `Config.max_budget` (default 120s) caps the foreground hold. The `python_exec` tool clamps to it and surfaces a note so a clamped long wait is resumed by polling `jobs['x']`, not silently lost.
- `Kernel._execute` runs the request as a shielded task and, on cancel, drains the reply with the lock held before re-raising, so the shared channel is never left mid-message.
- Guide: keep `budget` small and poll, rather than sitting on a huge foreground wait.

## Test
Extends the wedge smoke (`packages/mcp/default.nix`):
- cancelling an in-flight `python_exec` leaves the channel usable for the next call;
- the tool clamps a 600s budget to the 2s cap (returns within the cap, says it clamped).

Verified green by running the wedge test against this branch.

> [!NOTE]
> ### Fix `python_exec` shell channel desync on cancel and clamp foreground budget to `max_budget`
> - Wraps the `execute_interactive` call in an `asyncio` task and shields it from client cancellation in [`kernel.py`](https://github.com/indexable-inc/index/pull/878/files#diff-ced4fbbdf05b4f126e987804d73fdc2d811511b346e4985a9996ae7903683d1c); on cancel, the reply is drained under lock and `_interrupt()` is called if the drain times out before re-raising.
> - Adds `max_budget: float = 120.0` to [`Config`](https://github.com/indexable-inc/index/pull/878/files#diff-33f6f17aed79724ba72211bc8bf9dc1da1bbdb38de56be7ace76cc09837929ca) as a hard ceiling on a single `python_exec` foreground budget.
> - Clamps the requested budget to `Config.max_budget` in the [`python_exec` MCP tool](https://github.com/indexable-inc/index/pull/878/files#diff-5663cdf400300f377c9e4e2b021527ebf564c20364f79a77d25f888b294320f6) and appends a note to the response when clamping occurs.
> - Extends JOBS guidance in [`guide.py`](https://github.com/indexable-inc/index/pull/878/files#diff-03416e032f2f437bc1f5feeb6a43b54379221e5b079c2b2a98873b435daa8a76) to advise keeping budgets small and polling for long-running jobs.
> - Behavioral Change: cancelling an in-flight `python_exec` no longer desync the shared shell channel; budgets above `max_budget` are silently clamped server-side.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 687d4e6.</sup>
> 
>

### DIFF
--- a/packages/mcp/default.nix
+++ b/packages/mcp/default.nix
@@ -1079,7 +1079,7 @@ let
     # Result, jobs, the SIGUSR1/SIGUSR2 handlers) loads in the booted kernel,
     # exactly as the CLI wires it.
     os.environ["IPYTHONDIR"] = str(cli._prepare_ipython_startup(0))
-    config = Config(workdir=Path(tempfile.mkdtemp()), wedge_grace=1.0)
+    config = Config(workdir=Path(tempfile.mkdtemp()), wedge_grace=1.0, max_budget=2.0)
 
 
     async def main():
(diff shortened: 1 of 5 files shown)


andrewgazelka:
Applied: the cancel-drain now interrupts (SIGUSR2) when execute_interactive's timeout fires, so a synchronously-wedged cell is broken instead of waited out. Added a wedge-test case covering cancel of a sync-blocking cell. Commits 687d4e6a + the test.

PR #879, opened 2026-06-08
fix(mcp): stack a tuple/list of rich values instead of rasterizing it

andrewgazelka:
## What
`Result.of((repr_text, df))` coerced the whole tuple through `_as_frame_if_tabular`, which built a one-column frame and stringified the inner DataFrame into a cell. The human saw a 2-row frame of two reprs ("a text dataframe inside a dataframe") instead of the text plus the real table. Reported from:

```python
hits = fff.grep(...)
Result((repr(hits), hits.df))   # -> rasterized the inner df
```

## Change
A list/tuple that carries a rich element (a DataFrame, a figure, a nested `Result`) now routes to `_result_from_values`, so each element renders with its **own** view, stacked. A list/tuple of plain scalars (or of mappings) stays tabular, unchanged. `Result(repr(hits), hits.df)` (separate args) already worked; this makes the single-tuple form work too.

Also fixes a `richSmoke` assertion that is **already failing on `main`** (`nix build .#mcp.tests.richSmoke`): a list value renders as its one-column frame now, so the test asserts both values are present rather than the list's old `[1, 2, 3]` repr.

## Test
`packages/mcp/default.nix` rich smoke: a `(text, df)` tuple renders one real table plus the text (not a stringified frame); a plain scalar list stays one table.

> [!NOTE]
> ### Fix `Result.of` to stack list/tuple of rich values instead of rasterizing to a table
> - `Result.of` in [runtime.py](https://github.com/indexable-inc/index/pull/879/files#diff-a321ac036dbe5148143f01e84f705e2c20bda0e38f193e74b06c66889de27b95) now detects when the input is a list/tuple containing at least one rich element (Result, DataFrame, or displayable/figure) and stacks each element's own view via `_result_from_values` instead of coercing the whole sequence into a single table.
> - `Result.of` also gains an early-return path for an existing `Result` input that preserves `llm_images` and `user_html`, fixing a prior regression where those fields were lost.
> - Two new helpers, `_is_rich_element` and `_is_multi_rich`, encapsulate the detection logic.
> - Guide text in [guide.py](https://github.com/indexable-inc/index/pull/879/files#diff-03416e032f2f437bc1f5feeb6a43b54379221e5b079c2b2a98873b435daa8a76) is updated to document that multiple values can be passed as a list/tuple to stack views, and that DataFrames remain real tables rather than cells in an outer frame.
> - Behavioral Change: sequences with at least one rich element now render as stacked views; previously they were flattened into a single table.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5b2cdd8.</sup>
> 
>

### DIFF
--- a/packages/mcp/default.nix
+++ b/packages/mcp/default.nix
@@ -1177,6 +1177,16 @@ let
     runtime.install(ns)
     run = ns["__ix_run"]
 
+    # A tuple/list carrying a rich value (a DataFrame) renders each element with
+    # its own view, stacked, instead of stringifying the frame into a one-column
+    # table -- Result((repr_text, df)) shows the text AND the real table.
+    stacked = runtime.Result.of(("GrepResult: 0 matches", pl.DataFrame({"a": [1, 2]})))
+    assert stacked.user_html.count("<table") == 1, stacked.user_html[:200]
+    assert "GrepResult: 0 matches" in stacked.llm_result and "shape:" in stacked.llm_result, stacked.llm_result
+    # A plain list of scalars is still ONE table (not stacked), unchanged.
+    scalars = runtime.Result.of([1, 2, 3])
+    assert scalars.user_html.count("<table") == 1, scalars.user_html[:200]
+
 
     async def main():
         # A DataFrame result is stored with its text/html bundle.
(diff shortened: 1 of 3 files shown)


andrewgazelka:
Applied the suggestion: Result.of now copies an existing Result faithfully (keeping llm_images) before the multi-value path, so stacking a nested Result preserves its model image blocks. Added a rich-test case asserting images survive stacking.

PR #889, opened 2026-06-08
feat(mcp): decode sh stdout as JSON with Output.json/jsonl

andrewgazelka:
-

> [!NOTE]
> ### Add `Output.json()` and `Output.jsonl()` methods to decode shell stdout as JSON
> - Adds `Output.json()` to parse a command's stdout as a single JSON document, raising `ShellError` on non-zero exit instead of surfacing `JSONDecodeError` from error text.
> - Adds `Output.jsonl()` to parse stdout as newline-delimited JSON, returning a list of decoded values and skipping blank lines.
> - Both methods are tested in [packages/mcp/default.nix](https://github.com/indexable-inc/index/pull/889/files#diff-1620bd0b79c426472be159dd458dcedbb31de6b35b9507d98c31d4aed67e08db) covering success paths and error propagation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0ccc1ef.</sup>
> 
>

### DIFF
--- a/packages/mcp/default.nix
+++ b/packages/mcp/default.nix
@@ -1868,6 +1868,19 @@ let
         direct = await sh("printf hi", cwd=".")
         assert direct.ok and direct.text == "hi", repr(direct.text)
 
+        # Structured stdout decodes straight to Python (the polars on-ramp).
+        doc = await sh.sh("printf '%s' '{\"a\": 1, \"b\": [2, 3]}'", cwd=".")
+        assert doc.json() == {"a": 1, "b": [2, 3]}, doc.json()
+        rows = await sh.sh("printf '%s\\n%s\\n' '{\"n\": 1}' '{\"n\": 2}'", cwd=".")
+        assert rows.jsonl() == [{"n": 1}, {"n": 2}], rows.jsonl()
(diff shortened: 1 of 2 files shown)


andrewgazelka:
Landed onto `main` as bcd5a5cca417df65c5d892dc6adc8046f2cf70f0 (https://github.com/indexable-inc/index/commit/bcd5a5cca417df65c5d892dc6adc8046f2cf70f0). Built green via `nix build .#mcp.tests.shSmoke`.

On the AI reviewer's P1 (parsing the merged stderr): the `sh` module merges stdout and stderr at the PTY level by design (`stderr=asyncio.subprocess.STDOUT`), so there is no separate stdout to read; `.text`/`.stdout`/`.stderr` are all the one merged stream. `json()`/`jsonl()` parse `.text`, which is the right and only choice given that design, and they already guard the real footgun (a non-zero exit raises `ShellError` instead of a confusing `JSONDecodeError`). I tightened the docstrings to call out the merged-stream caveat and point at stderr redirection for chatty-on-success tools, and fixed the module-docstring example that called `.json()` where it meant `.jsonl()`.

(landed by an AI agent, Claude Fable 5)

PR #902, opened 2026-06-08
feat(mcp): expose an embeddable results API over the execution store

andrewgazelka:
## What

Make the `ix-mcp` execution store readable from outside the process, so a host that embeds `ix-mcp` (the room server runs Codex with `ix-mcp` as the agent's only tool) can fetch the agent's rich results and render them in its own chat UI.

- **`IX_MCP_DASHBOARD_PORT`** pins the data-API port (`cli._dashboard_port`). Today the port is a random free port announced only to the co-located dashboard, so an embedder can't address a specific instance. With the pin, the host sets the port when it launches `ix-mcp` and knows exactly where to poll.
- **`GET /api/jobs/<id>`** (backed by `store.get`) returns a single run by id, 404 otherwise. The model-facing `python_exec` reply already carries the job id, so the host fetches exactly the run behind a tool call and renders its tables/images/HTML inline rather than scanning the recent list.
- The dashboard module docstring now documents `/api/{jobs, jobs/<id>, resources, cells}` as the stable **embedding contract**.

## Why

First step toward merging the room server with `ix-mcp`: the agent does its work and *display* through `python_exec`/`Result`, and the room server is the surface that renders those results. That needs an addressable, by-run results API. This is decision-independent groundwork (true under any room-server topology) and changes no existing behavior: the dashboard still binds a free port when the env var is unset.

## Test

New `apiSmoke` (`packages/mcp/default.nix`, registered in `passthru.tests`): seeds the store, starts the server, and asserts the JSON routes (including the 404 and by-id paths) return the run's nbformat bundles, cells, and live resources. Built green via `nix build .#mcp.tests.apiSmoke`.

> [!NOTE]
> ### Expose a JSON results API over the MCP execution store
> - Adds `GET /api/jobs/{id}` to the dashboard server, returning a single execution as JSON with `code_html` syntax highlighting, or a 404 when the id is unknown.
> - Adds `store.get` in [store.py](https://github.com/indexable-inc/index/pull/902/files#diff-a4763dded9f44ec2a563e568a62d900b996c6a956b6de36462468f1d2b03b976) to fetch a single execution row by id with outputs and bindings parsed from JSON.
> - Introduces `_dashboard_port()` in [cli.py](https://github.com/indexable-inc/index/pull/902/files#diff-7ce741e8223d63fdea11625cb78abad266299445db0213d2355bc93306df9ac5) so the dashboard port can be pinned via `IX_MCP_DASHBOARD_PORT`, falling back to a free port when unset.
> - Adds an `apiSmoke` Nix check in [default.nix](https://github.com/indexable-inc/index/pull/902/files#diff-1620bd0b79c426472be159dd458dcedbb31de6b35b9507d98c31d4aed67e08db) that seeds the store, starts the server, and asserts `/api/jobs`, `/api/jobs/{id}`, `/api/cells`, and `/api/resources` return expected JSON structures.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4895d29.</sup>
> 
>

### DIFF
--- a/packages/mcp/default.nix
+++ b/packages/mcp/default.nix
@@ -923,6 +923,120 @@ let
         mkdir -p "$out"
       '';
 
+  # The read-only data API is also the embedding contract: a host (the room
+  # server runs `ix-mcp` as its agent's only tool) reads the agent's rich
+  # results back over HTTP and renders them in its own UI. Exercise that path
+  # in-process: seed the store, start the dashboard server, and assert the JSON
+  # routes (incl. the by-id lookup an embedder keys off the job id in a tool
+  # reply) return the run's nbformat output bundles, cells, and live resources.
+  apiTest = pkgs.writeText "ix-mcp-api-test.py" ''
+    import asyncio, tempfile
+    from pathlib import Path
+
+    import aiohttp
+
+    from ix_notebook_mcp import cli, dashboard, store
+    from ix_notebook_mcp.config import Config, set_config
+
+    # An embedder pins the data-API port so it knows where to reach this instance.
+    import os
+
+    os.environ["IX_MCP_DASHBOARD_PORT"] = "54321"
+    assert cli._dashboard_port() == 54321, cli._dashboard_port()
+    os.environ.pop("IX_MCP_DASHBOARD_PORT")
+    assert isinstance(cli._dashboard_port(), int)
+
+    tmp = Path(tempfile.mkdtemp())
+    store_path = tmp / "store.db"
+    conn = store.connect(store_path)
+    rich = [
+        {
+            "output_type": "execute_result",
+            "data": {
+                "text/plain": "shape: (1, 1)",
+                "text/html": "<table><tr><td>1</td></tr></table>",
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Landed onto `main` as d14f594cef76229ba9ff9242251551431f4dd7d3 (https://github.com/indexable-inc/index/commit/d14f594cef76229ba9ff9242251551431f4dd7d3).

Since this PR opened, the by-id results API it introduced (`GET /api/jobs/<id>` and `store.get`) already landed on `main` via the `feed` module, so to avoid a duplicate `job` handler and a second `store.get` I kept only the still-missing piece: the `IX_MCP_DASHBOARD_PORT` port pin (`_dashboard_port()`), plus your `apiSmoke` check (it now exercises main's `feed`-based routes end-to-end through `dashboard.start`, including the by-id 404 path, and validates the port pin). Built green via `nix build .#mcp.tests.apiSmoke`.

(landed by an AI agent, Claude Fable 5)

PR #1091, opened 2026-06-15
search-py: return a polars DataFrame from async semantic/grep/recent

### LINKED ISSUE #1089: search.* helpers return an awaitable but api()/signature give no signal (kind=func, no return annotation)
**Observation:** the `search.*` helpers (`semantic`, `grep`, `recent`) are sync functions that *return an awaitable*, but nothing short of the full docstring tells you so.

**Repro** (kernel):
```python
import inspect, search
inspect.iscoroutinefunction(search.grep)   # -> False  (api() therefore tags it kind=func)
inspect.signature(search.grep).return_annotation  # -> inspect._empty (no return type)
search.grep("nushell", top_k=1)            # -> _asyncio.Future, NOT a result
```

**Expected:** `api()` / the signature signals that the result must be awaited.
**Actual:** `kind=func` and no return annotation read as "just call it," so a caller writes `search.grep(...)` and gets `Future has no len()`. The truth lives only in `help(search.grep)` ("Returns an awaitable resolving to a list of dicts").

**Smallest change:** add a return annotation (`-> Awaitable[list[dict]]`) so `api()` surfaces it, or tag these awaitable in the `kind` column.

---
*(sent by an AI agent via Claude Code, Opus 4.8)*

### LINKED ISSUE #1090: search.* returns list[dict] but renders as a polars `shape:` table, implying DataFrame (inconsistent with fff/view)
**Observation:** `search.semantic/grep/recent` return `list[dict]`, but the kernel's rich display renders a list-of-dicts with a polars-style `shape: (N, M)` header, so it looks like a DataFrame.

**Repro** (kernel):
```python
r = await search.semantic("nushell", top_k=10)
type(r)          # -> list   (display showed "shape: (10, 12) | path:String, ...")
r.group_by(...)  # AttributeError: 'list' object has no attribute 'group_by'
```

This compounds because the sibling families `fff.grep` / `view.ls` *do* return polars frames, so the "everything is a frame" convention gets over-generalized to `search.*`.

**Expected:** the type is obvious from the output, or `search.*` matches `fff`/`view` and returns a polars frame.
**Actual:** a `list[dict]` is displayed with a polars `shape:` header, leading to `.group_by`/`.columns`/`["col"]` errors.

**Smallest change:** either return a polars frame from `search.*` (consistent with `fff`/`view`), or render lists without the frame-shaped header so the type isn't ambiguous.

---
*(sent by an AI agent via Claude Code, Opus 4.8)*

andrewgazelka:
## What

`search.semantic` / `grep` / `recent` now return a polars `DataFrame` (one row
per hit) instead of a `list[dict]`, and are real `async def` coroutines.

Closes #1089 — the binding read as a sync `func` with no return annotation, so
nothing signalled `await` was required (`api()` tagged it `func`; calling it
gave `Future has no len()`).

Closes #1090 — a `list[dict]` renders with a polars `shape: (N, M)` header, so
it looked like a DataFrame; `.group_by` / `df["col"]` then raised. `fff` and
`view` already return frames, so the inconsistency invited the wrong guess.

## How

A thin Python wrapper in `python/search/__init__.py` wraps each PyO3 binding:

```python
@functools.wraps(raw)
async def wrapper(*args, **kwargs):
    return _to_frame(await raw(*args, **kwargs))
```

- `functools.wraps` preserves the binding's full signature + docstring, so
  `help()` / `inspect.signature` are unchanged, while `iscoroutinefunction` is
  now True and `api()` tags them `async`.
- `_to_frame` enforces a stable, typed 14-column schema (the six always-present
  fields + provenance columns, null where absent), so `df["timestamp"]` and
  `df.group_by("source")` work even on an empty result.
- No Rust change: the `_search` cdylib still returns `list[Hit]`; this is the
  Python boundary layer the package already owns.

## Validation

Built the package's `__init__.py` against the shipped `_search.abi3.so` in a
clean interpreter:

- `semantic` / `grep` / `recent` are all `iscoroutinefunction=True` with their
  full signatures intact.
- each returns a `(N, 14)` DataFrame; empty results return a typed `(0, 14)`
  frame; ragged provenance keys union and null-fill.
- `df.group_by("source")`, `df["timestamp"]`, dtype enforcement all hold.

Docs updated: `search-py/README.md`, the `registry.py` module catalog blurb,
`agent-context/sections/18-searching.md`, and the `_search.pyi` header.

> [!NOTE]
> ### Return polars DataFrames from `search.semantic`, `search.grep`, and `search.recent`
> - Wraps the native PyO3 async functions with `_framed`, a decorator that converts `list[dict]` results to a polars DataFrame via `_to_frame`.
> - Enforces a stable column schema (`_COLUMNS`) with consistent ordering and correct dtypes; missing provenance columns are added as nulls.
> - Raises `ModuleNotFoundError` with a targeted message if polars is not installed.
> - Updates docs and MCP registry description to reflect the new DataFrame return type and show DataFrame operations like `.filter`, `.group_by`, and `.head`.
> - Behavioral Change: callers previously receiving `list[dict]` now receive a `polars.DataFrame`; polars becomes a required dependency at call time.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6075c38.</sup>
> 
>

### DIFF
--- a/agent-context/sections/18-searching.md
+++ b/agent-context/sections/18-searching.md
@@ -41,7 +41,9 @@ past.
 
 The shared store also indexes the whole fleet's history, not just code. From
 the index kernel: `import search`, then `await search.semantic("<task
-phrasing>", source=["claude_history"], top_k=5)`. Route by question type:
+phrasing>", source=["claude_history"], top_k=5)` — each verb is async and
+returns a polars frame (one row per hit; `.filter`/`.group_by`/`.head` it).
+Route by question type:
 `shell` answers "what is the command" for a few hundred tokens, `github`
 answers "why is it this way" with PR bodies and URLs, `claude_history` answers
 "how did someone do this". Pass `agentic=True` only for failure-shaped queries
--- a/packages/mcp/ix_notebook_mcp/registry.py
+++ b/packages/mcp/ix_notebook_mcp/registry.py
@@ -81,8 +81,9 @@ MODULES: tuple[Module, ...] = (
         "search",
         "meaning-based recall across the fleet corpus (code + agent/shell history): "
         "`await search.semantic(q, since='7d', compact=True)` / `grep(pattern)` / "
-        "`recent(source=['shell'], since='6h')` newest-first; hits carry timestamp/"
-        "user/host/session_id provenance",
+        "`recent(source=['shell'], since='6h')` newest-first. Each is async and "
+        "returns a polars frame (one row per hit, compose `.filter`/`.group_by`/"
+        "`.head`) with timestamp/user/host/session_id provenance columns",
         # Mirrors the resolution order owned by packages/mixedbread/src/auth.rs
         # (env key, else the `mgrep login` token), which the bundled module
         # reaches through search-py.
--- a/packages/search-py/README.md
+++ b/packages/search-py/README.md
@@ -12,26 +12,27 @@ boundary.
 ```python
 import search
 
-hits = await search.semantic("where is retry backoff configured")
-for hit in hits:
-    print(hit["path"], hit["score"], hit.get("timestamp"))
+df = await search.semantic("where is retry backoff configured")
+df.select("path", "score", "timestamp").head()
+df.group_by("source").len().sort("len", descending=True)
 
 # only Claude history, only my records, last two weeks, token-frugal
-hits = await search.semantic(
+df = await search.semantic(
     "deploy steps", source=["claude_history"], user=["andrew"],
     since="2w", compact=True,
 )
 
 # my shell commands of the last six hours, newest first
-rows = await search.recent(source=["shell"], user=["andrew"], since="6h")
+df = await search.recent(source=["shell"], user=["andrew"], since="6h")
 
-hits = await search.grep(r"fn \w+\(", source=["code"], repo="indexable-inc/index")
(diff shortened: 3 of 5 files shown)


andrewgazelka:
Force-merging despite a red `flake-check`. The failures are pre-existing and unrelated to this pure-Python `search` change: `main` itself is red on all 6 most recent pushes, and the failing attrs are `rust-mcp.browserSmoke`/`browserVdomSmoke`/`vdomPropertiesSmoke` (chromium-in-sandbox flakes: ECONNREFUSED / TargetClosed), `image-development-base`/`image-symphony-codex` (docker image builds), and `rust-mcp.fffBundled` (an `fff` test) — none touch `search`.

Per the force-merge safety gate, I built the exact head SHA locally on aarch64-darwin and ran the artifact:

```
$ nix build .#packages.aarch64-darwin.mcp   # EXIT=0 -> /nix/store/69glsz9msrsjbiqnmp0f17j05syvljz7-ix-mcp
$ <built python> import search; ...
version: 0.1.0
semantic async= True / grep async= True / recent async= True
shapes: (5, 14) (4, 14) (3, 14)
ARTIFACT OK
```

(sent by an AI agent via Claude Code, Opus 4.8)

PR #1117, opened 2026-06-16
feat(ix-fleet): declarative per-VM secret refs

harivansh-afk:
## What

Fleet authors can now attach user-store secrets to a VM **declaratively**, instead of only imperatively via `ix new --secret NAME`.

```nix
web = {
  deployment = {
    secrets = [ "GH_TOKEN" "DATABASE_URL" ];  # names from `ix secret set`
    noDefaultSecrets = true;                   # opt out of --default auto-attach
  };
};
# deployment.secrets at the fleet level unions into every node (like l7ProxyPorts)
```

## Why

The user secret store (ENG-1779) is declarative-ready end to end - `build_commit_from_oci` takes `secrets`/`no_default_secrets`, `ix new --secret` uses it, and the Python SDK `create(secrets=, no_default_secrets=)` exposes it. The only missing surface was the fleet plan: `FleetNode` had no way to reference store secrets and `create_node` never passed them.

(`FleetPlan.secrets` already existed but is a different mechanism - the vaultwarden/runtime-directory provider for infra secrets, not the per-user store.)

## Changes

- **`FleetNode`** gains `secrets` (names from the per-user store) + `noDefaultSecrets`; `create_node` forwards them to `client.create`.
- **`lib/image/fleet.nix`** surfaces `deployment.secrets`/`noDefaultSecrets`, unioning names across fleet-wide and per-node layers, and emits them into the plan JSON.
- **`verify_secrets_available`** preflights `up`/`switch`/`replace`/`bootstrap`, failing fast on missing references before any image push (mirrors the `ix secret check` deploy-time bridge).

References only, never values: plaintext stays in the store, not the plan.

## Tests

- New Python unit tests (model round-trip + preflight pass/fail/dry-run/no-ref).
- New Nix assertion in `tests/default.nix` (fleet-wide + node-level union, opt-out).
- `dryRunUp` plan exercises the field through the dry-run create path.

Verified locally: Python tests pass against a stubbed SDK; the Nix generator pure-evaluates to the expected per-node `secrets`/`noDefaultSecrets`. Full NixOS eval / `dryRunUp` run in x86_64 CI.

> [!NOTE]
> ### Add declarative per-VM secret refs and preflight validation to ix-fleet
> - Adds `secrets: list[str]` and `noDefaultSecrets: bool` fields to the `FleetNode` pydantic model, passed through to `client().create` at VM creation time.
> - Introduces `verify_secrets_available`, an async preflight that compares referenced secret names against `client().list_secrets()` and raises `RuntimeError` for any missing; called before work begins in `cmd_up`, `cmd_switch`, `cmd_replace`, and `cmd_bootstrap`.
> - Fleet-level and per-node secret refs are unioned in [fleet.nix](https://github.com/indexable-inc/index/pull/1117/files#diff-9836ea3ef1875452d04d6213563ea9b0df81e37a277277836b0cb9df0ced1fa0) and propagated through the plan; node-level `noDefaultSecrets` is preserved independently.
> - Adds build-time surface checks to [ix-sdk-python/default.nix](https://github.com/indexable-inc/index/pull/1117/files#diff-80f46c8d1744ebc070c811611b94e45d792b24cdeeae53fcaf78e4db15945032) asserting `list_secrets` and secret kwargs exist on the SDK client.
> - Risk: all four commands now abort before any VM work if a referenced secret is absent from the user store (no change in dry-run mode).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9cef4a6.</sup>
> 
>

### DIFF
--- a/lib/image/fleet.nix
+++ b/lib/image/fleet.nix
@@ -53,6 +53,9 @@ let
     // {
       env = lib.mergeAttrsList (map (part: part.env or { }) parts);
       l7ProxyPorts = lib.unique (lib.concatMap (part: part.l7ProxyPorts or [ ]) parts);
+      # User-store secret names union across defaults/fleet/node layers (like
+      # l7ProxyPorts), so a fleet-wide default attaches alongside per-node refs.
+      secrets = lib.unique (lib.concatMap (part: part.secrets or [ ]) parts);
     };
 
   # Every deployment key the plan consumes. `deployment` is a plain attrset
@@ -69,8 +72,10 @@ let
     "env"
     "ipv4"
     "l7ProxyPorts"
+    "noDefaultSecrets"
     "recreateOnUp"
     "region"
+    "secrets"
     "snapshot"
     "switch"
   ];
@@ -289,6 +294,10 @@ let
       groups = nodeGroups;
       inherit (deploy) env;
       inherit (deploy) l7ProxyPorts;
+      # Per-VM user-store secret references, attached at create like
+      # `ix new --secret NAME`; `ix-fleet` verifies they exist before deploying.
+      secrets = deploy.secrets or [ ];
+      noDefaultSecrets = deploy.noDefaultSecrets or false;
       dependsOn = expandedDependencies.${name};
       healthChecks = planHealthChecks config;
     }
--- a/packages/ix-fleet/default.nix
+++ b/packages/ix-fleet/default.nix
@@ -38,6 +38,10 @@ let
       region = "us-west-1";
       ipv4 = false;
       snapshot = false;
+      # Exercise the declarative per-VM secret refs through the dry-run create
+      # path (no live call, so the names need not exist in any store here).
+      secrets = [ "GH_TOKEN" ];
+      noDefaultSecrets = true;
     };
   };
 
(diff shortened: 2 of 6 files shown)


harivansh-afk:
## CI triage

**AI review (2× P1) — false positive, but addressed.** Both findings claim the packaged Python SDK lacks `Client.list_secrets` / the `secrets`/`no_default_secrets` create kwargs. They were read against the stale checked-in reference copy at `sdk/python/ix_sdk/`, **not** the prebuilt wheel that ix-fleet actually imports (injected from `packages/ix-sdk-python`, fetched from R2).

I downloaded and inspected that wheel (`ix_sdk-0.1.0-cp313-abi3-manylinux_2_34_x86_64.whl`) directly. Its `ix_sdk/__init__.py` exposes exactly the surface this PR uses:
- `async def create(..., secrets: list[str] | None = None, no_default_secrets: bool = False)` (L977-978)
- `create_with_progress(..., secrets=, no_default_secrets=)` (L1005-1006)
- `async def list_secrets(self) -> list[UserSecret]` (L1063)

The wheel sync (#1075, 2026-06-13) post-dates the SDK secrets commit (#4546, 2026-06-09), consistent with this.

To convert that from inference into an enforced guarantee — and to address the reviewer's "or asserted" point — `9cef4a6` extends the existing `ix-sdk-python` import proof to assert `Client.list_secrets` plus the `secrets`/`no_default_secrets` kwargs on `create`/`create_with_progress`. A stale wheel now fails that check instead of failing ix-fleet at deploy.

**`flake-check` — pre-existing/environmental, not from this PR.** It is also red on `main` right now. The failing derivations are GTK/webkit system-lib build scripts (`atk-sys`, `cairo-sys-rs`, `gdk-sys`, `gtk-sys`, `webkit2gtk-sys`, `tao`, `wry`) failing pkg-config in the build sandbox. This PR adds zero Rust/Cargo/GUI dependencies; `ix-fleet.dryRunUp` (its own check) is wheel-based and independent of that chain.

PR #1139, opened 2026-06-17
Elixir set-theoretic type gate (astlog rules + compile check) + hive package

andrewgazelka:
## What

Brings Elixir under the same "type check by default" discipline the repo already has for Python, in two layers, plus a small reference package and the engine support to express the rules.

```mermaid
flowchart LR
  A[Elixir lib/*.ex] --> B[nix run .#lint: astlog-elixir]
  A --> C["package check: mix compile --warnings-as-errors"]
  B -->|defstruct needs @type, public def needs @spec| D[lint gate]
  C -->|1.18 set-theoretic checker| E[CI flake check]
```

### Lint layer — `astlog-rules/elixir.astlog`
Two rules, run over **every package's `lib/`** by `nix run .#lint` (`astlog-elixir` stage):
- `defstruct-needs-type`: a module with a `defstruct` must declare a `@type`.
- `public-def-needs-spec`: a public `def` must have an adjacent `@spec`; behaviour callbacks marked `@impl` are exempt.

The only scoping is to `lib/` — `mix.exs` build functions and `test/` helpers are not the type-checked surface. Good/bad fixtures under `astlog-rules/tests/` are checked by the existing `astlog-rules` self-test (extended for `elixir.astlog`).

### Build layer
Each Elixir package's `mix compile --warnings-as-errors` (which runs Elixir 1.18's set-theoretic type checker) is a sandboxed flake check — `hive-elixir`, mirroring `symphony-elixir`.

### astlog engine
- New `no-prev-sibling` builtin: the sibling-scoped analogue of `no-descendant`, so a rule can require an annotation directly above a definition.
- `is_named` flag on the node table (skip anonymous tokens when walking siblings).
- Sound by design: checks the immediately-preceding sibling, never hides an unspec'd def. Multi-clause functions are judged per clause (spec each, or `# astlog-ignore` continuations). Unit test added.

### New package — `packages/andrewgazelka/hive`
A small fully-connected GenServer agent mesh (moved out of a standalone checkout), the reference user of the gate: typed `Hive.Agent.State` struct, `@spec` on every public function, both gates green. `nix run .#hive` runs the demo.

### symphony
Added `@spec`/`@impl` to the public API its `lib/` exposes so it passes the new repo-wide gate. **No logic changed** — annotation lines only.

## Verification (local, darwin)
- `nix run .#lint` → 6/6 green (incl. `astlog-elixir` repo-wide).
[...]

PR #1143, opened 2026-06-17
astlog: stratified negation; make Elixir type-gate sound

andrewgazelka:
## What & why

PR #1139 shipped the Elixir type-discipline lint with three documented blind spots, all rooted in astlog v0 having **no negation** — a rule couldn't say "there is no `@spec` *matching this def's name*." This adds **stratified negation** to the engine (a reusable capability, listed as a v0 gap in the README) and rewrites the Elixir rules to be name/structure-matched and sound.

```mermaid
flowchart TD
  A["(not (relation ...))"] --> B[range-restriction: vars bound by a positive atom]
  A --> C[stratification: no relation negates itself]
  C --> D[evaluate stratum by stratum]
  D --> E[negated relation is final before it is read]
```

## Engine — `packages/code/astlog/core`
- New `(not (relation args...))` body atom. Rejected if a negated variable isn't bound by a positive atom (`UnsafeNegation`) or if a relation depends on itself through negation (`UnstratifiableProgram`). `fixpoint` now runs stratum by stratum.
- Flipped `no-attached-sibling` → positive `attached-sibling`; absence is expressed with `(not ...)`. Kept #1139's bounded-walk perf.
- Unit tests: name-join negation, unsafe-negation rejected, recursion rejected, annotation-block scan.

## Rules — `astlog-rules/elixir.astlog`
- **`public-def-needs-spec`**: flag a public `def` unless its module has a `@spec`, or an `@impl`-annotated def, with the **same function name**. Fixes blind spot #1 (a spec for a *different* function no longer exempts a def) and removes the multi-clause `astlog-ignore` clutter (one spec/impl covers every clause). `@impl` callbacks exempt.
- **`defstruct-needs-type`**: matches the real `@type`/`@typep`/`@opaque` attribute, not any identifier named `type` (blind spot #2).
- **Residual (documented):** matching is by name, not arity — astlog has no aggregation to count args. The real type gate (`mix compile --warnings-as-errors`) is unaffected.

## Self-test — `lib/per-system.nix`
Dropped the "every `(rule ...)` must back a `(lint ...)`" check — helper relations are legitimate Datalog (joins/negation need them), and this invariant is what bit #1139. The meaningful checks remain: every lint has good/bad fixtures that fire/stay-clean, every fixture backs a lint, and `astlog` rejects a lint naming an undefined relation at parse.

## Cleanup
Removed the 49 now-redundant `# astlog-ignore: public-def-needs-spec` comments across symphony + hive `lib/`.

## Verified locally (darwin)
- `nix build .#astlog` builds; negation + every rule case checked end-to-end with the binary; self-test replica passes (bad fires, good clean for both lints).
- `nix run .#lint` → 6/6 green. `nix build .#symphony.passthru.tests.elixir .#hive.passthru.tests.elixir` → green (compile -Werror, format, credo, test). 0 findings across all `lib/`.
- The `astlog-rules` self-test + `*-elixir` checks are `x86_64-linux`-only, so CI builds those.

Reviewers from astlog: the stratification is in `program.rs::strata`; the negation eval arm is in `eval.rs::solve`.

🤖 Opened by an AI agent (Claude Code / claude-opus-4-8) on behalf of @andrewgazelka.

> [!NOTE]
> ### Add stratified negation to astlog and fix Elixir type-gate soundness
> - Adds `(not (relation ...))` syntax to the astlog rule language, with parse-time safety checks (unsafe negation, unstratifiable cycles) and stratum-ordered evaluation in [eval.rs](https://github.com/indexable-inc/index/pull/1143/files#diff-582d2ea95f2cadbafa86621c62413054f21723e09bfe15a824bbcefcdf51a615) and [program.rs](https://github.com/indexable-inc/index/pull/1143/files#diff-49a2029f4d0a9f0a74b390eadfaf4fb3ae4c07fcb6da07928ba22eeb4bf4147d).
> - Rewrites the `defstruct-needs-type` lint in [elixir.astlog](https://github.com/indexable-inc/index/pull/1143/files#diff-7c3885af2a45aaf196cbe344831d5b87cf5d8d6554a44d9d747e083e0841e1a0) to match only true `@type`/`@typep`/`@opaque` attributes scoped to the nearest enclosing module, replacing a heuristic that matched any identifier named `type`.
> - Rewrites the `public-def-needs-spec` lint to require a same-named `@spec` (or `@impl`) in the same nearest module, so a spec for a different function or in an ancestor module no longer exempts a def; multi-clause functions are covered by name.
> - Renames the `no-attached-sibling` builtin to `attached-sibling` and inverts its polarity; callers must now use `(not (attached-sibling ...))` for absence checks.
> - Removes ~20 `# astlog-ignore: public-def-needs-spec` suppression comments from Elixir source files now that multi-clause defs are correctly handled by name.
> - Risk: any external `.astlog` file using `no-attached-sibling` will break; the builtin no longer exists.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2fe6b63.</sup>
> 
>

### DIFF
--- a/astlog-rules/elixir.astlog
+++ b/astlog-rules/elixir.astlog
@@ -29,62 +29,88 @@
   (match elixir "(unary_operator (call (identifier) @kw (arguments (binary_operator left: (call (identifier) @n))))) @s")
   (text kw "spec"))
 
-;; --- defstruct-needs-type: a struct module must declare a @type ---
-;; Matches the actual `@type`/`@typep`/`@opaque` attribute (a `@` unary_operator
-;; over a call to `type`/...), not merely any identifier with text `type`.
+;; The actual `@type`/`@typep`/`@opaque` attribute (a `@` unary_operator over a
+;; call to `type`/...), not merely any identifier with text `type`.
 (rule (type-attr a)
   (match elixir "(unary_operator (call (identifier) @ti)) @a")
   (text-match ti "^(type|typep|opaque)$"))
 (rule (defstruct-call c)
   (match elixir "(call (identifier) @i) @c")
   (text i "defstruct"))
+
+;; --- nearest enclosing module ---
+;; A spec/impl/struct/type belongs to its NEAREST enclosing `defmodule`, not every
+;; ancestor module: otherwise a nested module's `@spec foo` would also count for
+;; the outer module and wrongly exempt an unspec'd `Outer.foo`. `nearest-mod x m`
+;; holds when `m` is a defmodule ancestor of `x` with no defmodule between them.
+(rule (mod-candidate x)
+  (def-name x n))
+(rule (mod-candidate x)
+  (spec-name x n))
+(rule (mod-candidate x)
+  (defstruct-call x))
+(rule (mod-candidate x)
+  (type-attr x))
+;; `m` is an ancestor module of `x` but a closer module `mid` sits in between.
+(rule (mod-shadowed x m)
+  (mod-candidate x)
+  (defmodule m)
+  (ancestor m x)
+  (defmodule mid)
+  (ancestor m mid)
+  (ancestor mid x))
+(rule (nearest-mod x m)
+  (mod-candidate x)
+  (defmodule m)
+  (ancestor m x)
+  (not (mod-shadowed x m)))
+
+;; --- defstruct-needs-type: a struct module must declare a @type ---
 (rule (module-has-type m)
   (type-attr a)
-  (defmodule m)
-  (ancestor m a))
-(rule (defstruct-needs-type m)
-  (defmodule m)
+  (nearest-mod a m))
+(rule (module-has-struct m)
   (defstruct-call c)
-  (ancestor m c)
+  (nearest-mod c m))
+(rule (defstruct-needs-type m)
+  (module-has-struct m)
   (not (module-has-type m)))
 (lint defstruct-needs-type error
   "module defines a struct but no @type; add a `@type t` for it so the 1.18 type checker can type its fields")
 
 ;; --- public-def-needs-spec: a public def needs a @spec (or @impl) ---
-;; A def is flagged unless its module has a @spec, or an @impl-annotated def, with
-;; the SAME function name. Name-matching makes one `@spec`/`@impl` cover every
-;; clause of a multi-clause function, and prevents a spec for a different function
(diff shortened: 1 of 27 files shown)


andrewgazelka:
Adversarial review run; 5 confirmed findings addressed in `aebf3333`:

**Fixed (correctness blockers):**
- **Order-dependent negation safety** — `check_negation_safety` was order-independent while the evaluator runs left-to-right, so a `(not …)` placed *before* its binding atom passed the check but bound its var existentially (a wildcard) instead of filtering. Now the check walks the body in order and requires each negated var to be bound by a positive atom that **precedes** the negation. Test added (negation-before-binding rejected).
- **Rewrite bodies unchecked** — the safety check ran only on rule bodies. Generalized over `&[BodyAtom]` and applied to rewrites too. Test added.

**Fixed (maintainability):**
- `UnstratifiableProgram` blamed rule #1; now reports a relation actually on the negation cycle.

**Declined (judgment calls):**
- **perf: corpus-wide module×def join.** Real, but at the repo's ~78 Elixir `lib/` files it's 0.25s, and the obvious per-file-scan "fix" is **18× slower** there (process startup), so it's the wrong fix. The pre-existing rules were already quadratic. Proper fix is engine-level relation indexing / same-file scoping — filed as #1144 rather than bundled here.
- **perf: per-row `Env` clone in the negation arm** is the engine's universal join primitive (positive joins clone identically); optimizing only negation would be inconsistent.

Verified locally: astlog builds; negation order-safety enforced; `nix run .#lint` green; symphony + hive `lib/` 0 findings; self-test replica passes.

🤖 (Claude Code / claude-opus-4-8, on behalf of @andrewgazelka)

andrewgazelka:
Addressed the P2 in `2fe6b631`: spec/impl/type membership is now scoped to the **nearest enclosing `defmodule`** (new reusable `nearest-mod` relation via negation of `mod-shadowed`), so a nested module's `@spec`/`@impl`/`@type` no longer exempts an outer same-named def. Verified with nested-module cases; symphony+hive still 0 findings (~0.33s). (The P1 negation-ordering issue was fixed in the prior commit.)

The `ix-mcp-*-smoke` flake-check failures don't touch any code in this PR (astlog/rules only) and 5 unrelated smoke variants failed together — re-running on this fresh commit to confirm they were flaky kernel-boot infra.

🤖 (Claude Code / claude-opus-4-8)

PR #1313, opened 2026-06-17
mcp/iphone: reap the privileged tunneld on cleanup

andrewgazelka:
Fast-follow to #1311. That PR merged the version before the final cleanup fix landed, so `main`'s `iphone` helper still:

- leaks a **root** `tunneld` on the `start_tunneld` readiness timeout (raises "not ready in 20s" without stopping the daemon — the P2 the #1311 AI review flagged), and
- uses `proc.terminate()` in `stop_tunneld`, which **EPERMs** because the daemon runs as root (spawned via `sudo`).

This routes termination through `sudo -n kill` on the process group (the daemon has its own session via `start_new_session=True`, so its pid is the pgid), and escalates **SIGTERM → SIGKILL** so a stuck daemon is never left running as root.

Verified live on the iPhone 16 / iOS 26.5.1: `start_tunneld(sudo=True)` → 1 tunnel → `stop_tunneld()` → 0 tunnels, no leftover process.

🤖 Generated with Claude Code

> [!NOTE]
> ### Reap the privileged tunneld process group via sudo on cleanup
> - Adds `_group_signal` and `_sudo_kill` helpers in [`__init__.py`](https://github.com/indexable-inc/index/pull/1313/files#diff-f92c2ac3b6f1cee30bc3c534269f16357522a2d23c4c9e1e984a8f32030227b7) that use `sudo -n kill` to send signals to the root-owned tunneld process group, with SIGTERM→SIGKILL escalation and a 5s wait between attempts.
> - `stop_tunneld` now calls `_sudo_kill` instead of the unprivileged `proc.terminate()`/`proc.kill()`, which previously had no effect on a root-owned process.
> - On failed startup, `start_tunneld` attempts `_sudo_kill` in the `finally` block (suppressing `IphoneError` to avoid masking the original error) and retains `_tunneld_proc` if the process has not yet exited.
> - Risk: if `sudo` is not configured to allow passwordless kill, `_sudo_kill` raises `IphoneError` rather than silently failing.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c73fb5d.</sup>
> 
>

### DIFF
--- a/packages/mcp/src/iphone/iphone/__init__.py
+++ b/packages/mcp/src/iphone/iphone/__init__.py
@@ -35,6 +35,7 @@ one is attached (``devices()`` lists them). Targeting is done through the CLI's
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import json
 import os
 import re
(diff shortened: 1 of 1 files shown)


andrewgazelka:
Superseded by #1316, which folds in this tunneld cleanup plus the full UI-automation rewrite.

PR #1327, opened 2026-06-18
ix-windows: floating blurred overlays that auto-size to content

andrewgazelka:
## What

Reworks `ix-windows` from tiled borderless webview windows into floating, blurred **overlay** cards that auto-size to their content. No tiling, no layout manager.

- **Blur behind.** Each window paints its transparent `wry` webview on top of a native `NSVisualEffectView` (behind-window blur, `HUDWindow` material), inserted as the content view's first subview with a rounded, shadowed layer. The rendered HTML sits on top.
- **Auto-size to content.** A page `ResizeObserver` measures the `#ix-root` panel and posts its pixel size over `wry`'s IPC channel; the new `UserEvent::Resize` drives `WindowManager::resize`, which fits the OS window (clamped to the monitor, 1px hysteresis to break any reflow loop). The panel is `inline-block`/`max-width` so its intrinsic size doesn't depend on window width.
- **Floating overlay.** Windows are transparent + borderless + always-on-top, and join all spaces / float over fullscreen apps (`NSWindowCollectionBehavior`).

## Why

The previous behavior relied on an external tiling WM (aerospace/yabai rules) to lay windows out. This makes each resource a self-contained, self-sizing HUD overlay instead.

## Validation

- `nix build .#ix-windows` passes (darwin).
- New `parse_size` unit test; existing `shell`/`escape_text` tests retained. Window/webview/blur paths need a display and are exercised manually.
- Docs + README updated (`docs/ix-windows/overview.md`).

🤖 Drafted with Claude Code (Opus).

> [!NOTE]
> ### Add floating blurred overlay windows with auto-size to content in ix-windows
> - Windows are now always-on-top transparent overlays instead of borderless square windows, with initial size reduced to 480x300 and position cascade adjusted.
> - A `ResizeObserver`-based script ([MEASURE_JS](https://github.com/indexable-inc/index/pull/1327/files#diff-1e1abdcf7151daee9cdb984d29eca28bda3ded2eaa1b78937674b8156ef55a43)) measures the `#ix-root` panel and reports pixel dimensions to the native layer via wry IPC, driving `WindowManager::resize` to clamp and apply the new size against monitor bounds.
> - On macOS, `install_blur()` uses `objc2-app-kit` to insert an `NSVisualEffectView` beneath the webview and sets collection behavior to float across spaces and over fullscreen apps.
> - A new `UserEvent` enum wraps both producer stream events and `Resize` reports; `main.rs` forwards producer events as `UserEvent::Producer` and dispatches `UserEvent::Resize` to `WindowManager::resize`.
> - `parse_size` strictly validates IPC size messages, rejecting non-finite or non-positive values to avoid panics during clamping.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1895a3e.</sup>
> 
>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4112,6 +4112,7 @@ dependencies = [
  "clap",
  "dashboard-core",
  "objc2",
+ "objc2-app-kit",
  "objc2-foundation",
  "objc2-web-kit",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -204,6 +204,7 @@ nix = "0.31"
 nix-web-monitor-parser = { path = "packages/nix/nix-web-monitor/parser" }
 numpy = "0.28"
 objc2 = "0.6"
+objc2-app-kit = "0.3"
 objc2-foundation = "0.3"
 objc2-web-kit = "0.3"
 object_store = { version = "0.13", features = ["aws"] }
--- a/docs/index.md
+++ b/docs/index.md
@@ -50,7 +50,7 @@ From-source documentation for the packages in the `index` repo (a shared, open-s
 | [ix-dev-diagnose](ix-dev-diagnose/overview.md) | `packages/ix-dev-diagnose` probes `https://ix.dev/` (or any HTTPS URL) from the caller's network path and writes a single JSON diagnostic capturing every layer of the request: system DNS... |
 | [ix-fleet](ix-fleet/overview.md) | `packages/ix-fleet` renders and executes declarative **fleet plans**: a single JSON document describes a set of remote ix VMs (nodes) and their images, NixOS switch targets, east-west gro... |
 | [ix-sdk-python](ix-sdk-python/overview.md) | `packages/ix-sdk-python` is the Nix package that makes the precompiled Python SDK bindings available in-repo as `pkgs.ix-sdk-python` / `nix build .#ix-sdk-python`. |
-| [ix-windows](ix-windows/overview.md) | `packages/ix-windows` renders each live MCP resource as its own borderless, square, ghostty-styled native webview window. |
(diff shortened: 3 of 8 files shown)


andrewgazelka:
### Code review resolved

Ran an adversarial review pass. Changes pushed:

- **C1 (blocker) fixed** — `parse_size` now rejects non-finite/non-positive values. The IPC body is attacker-controlled (any resource's verbatim HTML can call `window.ipc.postMessage`), and `f64::clamp` *panics* on `NaN`, which would have aborted the event loop and killed every overlay. Added a unit test for `NaN`/`inf`/`1e400`/negative/zero.
- **M1 (doc) fixed** — `resize`'s comment now credits the page-side dedup (`MEASURE_JS` only posts on an actual size change; the panel's intrinsic `inline-block`/`max-width` size doesn't depend on window width) as the primary loop-breaker; the 1px Rust guard only suppresses sub-pixel jitter.
- **C2 (blur tracking)** — kept the flexible-autoresizing-mask approach and documented why it's sufficient: an `NSWindow` always resizes its content view to the content rect and `contentView.autoresizesSubviews` defaults on, so the blur tracks every `set_inner_size` with no explicit handler — the same mechanism that keeps wry's own webview filling the window. Worth a visual confirm on a ProMotion display when convenient, but no code path relies on an unverified assumption beyond standard AppKit.

Confirmed clean (no action) by the reviewer: retain balance in `install_blur`, subview ordering (blur below webview), IPC-handler thread-safety (`EventLoopProxy` is `Send+Sync`), and `Resize`-for-closed-window lifecycle.

`nix build .#ix-windows` passes.

🤖 via Claude Code (Opus)

PR #1339, opened 2026-06-18
ix-windows: close button, edge-to-edge content, thinner scrollbars

andrewgazelka:
## What

Make the `ix-windows` overlay cards manageable per the request: close them when no longer needed, content flush to the edge, more transparent, much thinner scrollbars.

- **Close button** — borderless overlays had no native X, so `CloseRequested` never fired and a card could only vanish when its resource did. Adds a floating `×` (top-right, faint base opacity, brighter on hover) wired through a new `UserEvent::Close` + a `"close"` IPC body → `WindowManager::window_closed`. Reuses the existing *dismissed* lifecycle, so closing a still-live resource keeps it shut until it re-registers.
- **Edge-to-edge content** — card padding dropped to `0`; `overflow:hidden` clips the iframe to the rounded corners.
- **Stays movable with zero padding** — content fills the card, so the old grab-the-padding drag is gone. Window now moves via **Cmd+drag anywhere** (relayed from the sandboxed iframe) or a press on any bare background.
- **More transparent** — card tint alpha `0.30 → 0.20`.
- **Thinner scrollbars** — `10px → 4px`.

## Why the close button is always faintly visible

The card content is a sandboxed (`allow-scripts`, opaque-origin) iframe, and hovering inside it does **not** set `:hover` on the parent card. A pure reveal-on-card-hover X would only show while the cursor sat on the button itself — undiscoverable. So the X has a faint always-on base opacity.

## IPC / trust boundary

```mermaid
flowchart LR
  P[producer HTML<br/>sandboxed iframe] -->|postMessage ixsize/ixdrag| O[trusted outer doc]
  O -->|window.ipc: size / drag / close| R[wry IPC handler<br/>captures own window id]
  R --> E[UserEvent::Resize/Drag/Close]
  E --> M[WindowManager]
```

Each webview's IPC handler captures its own window `id`, and the outer doc gates iframe messages on `event.source === frame.contentWindow`, so a producer can only resize/move/close **its own** card — same authority it already had for resize.

## Validation

- `nix build .#ix-windows` ✅
- Ran the real binary on macOS: confirmed edge-to-edge content, faint X revealing on hover, thin 4px scrollbars (screenshots in review).
- Deterministic Chromium test of the exact `shell()` HTML: `#ix-root`/`#ix-content` padding `0px`; card bg `rgba(30,30,46,0.2)`; X base opacity `0.4`; clicking X posts `close`; Cmd-drag relays `drag`.
- `code-reviewer` pass: no security blockers (cross-card forge not possible). Two correctness findings fixed — Cmd-click on producer content no longer has its default suppressed (only bare-background drag does), and the zero-size gate now matches the Rust `parse_size` `<= 0` contract.

🤖 Generated with Claude Code (Opus 4.8)

> [!NOTE]
> ### Add floating close button, edge-to-edge content, and thinner scrollbars to ix-windows
> - Adds a floating close button to each window's outer shell document. Each window gets a per-window random UUID token via `close_token()`; the close button posts `close:{token}` over IPC, which the host validates before emitting `UserEvent::Close` and calling `window_closed`.
> - Enables Cmd+drag from anywhere inside the iframe (or bare background drag) to move windows by relaying `ixdrag` messages from the inner frame to the outer shell.
> - Removes padding from the outer shell so iframe content fills the window edge-to-edge; zero-size content reports are clamped to 1px to prevent stale window sizes.
> - Reduces scrollbar width/height from 10px to 4px in the inner CSS.
> - Removes macOS layer-based corner rounding so windows are now square-cornered.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 45500ea.</sup>
> 
>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4118,6 +4118,7 @@ dependencies = [
  "serde_json",
  "tao",
  "tokio",
+ "uuid",
  "wry",
 ]
 
--- a/packages/ix-windows/Cargo.toml
+++ b/packages/ix-windows/Cargo.toml
@@ -22,6 +22,9 @@ dashboard-core.workspace = true
 serde_json.workspace = true
 tao.workspace = true
 tokio = { workspace = true, features = ["rt-multi-thread", "sync", "net", "io-util", "time"] }
+# `v4` for the per-window close token: a 128-bit OS-random value, unguessable by
+# sandboxed producer script (see `close_token`).
+uuid = { workspace = true, features = ["v4"] }
 wry.workspace = true
 
 # Native macOS overlay styling via objc2: the `NSVisualEffectView` blur behind
--- a/packages/ix-windows/README.md
+++ b/packages/ix-windows/README.md
@@ -30,10 +30,15 @@ tiling, no layout manager.
   content grows.
 - **Floating across spaces.** The window is always-on-top and joins all spaces /
   floats over fullscreen apps (`NSWindowCollectionBehavior`).
-- **Move, don't resize.** The card is borderless, so there is no title bar: drag
-  it by its **chrome** (the padding/background around the content) to move it. The
-  window is not user-resizable -- its size is owned by the content (auto-fit), so a
-  manual resize would just fight the next content report.
+- **Move with Cmd+drag.** The card is borderless and runs content flush to its
+  edge (no padding), so there is no chrome to grab. Hold **Cmd** and drag anywhere
+  -- even over content -- to move it (an ordinary Cmd+click on content still works);
+  a drag on any bare background area moves it too. The window is not user-resizable
+  -- its size is owned by the content (auto-fit), so a manual resize would just
+  fight the next content report.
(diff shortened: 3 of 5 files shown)


andrewgazelka:
All P1/P2 review findings (off-screen close button, dropped zero-height resize, capture-phase drag, forgeable close IPC) are addressed in d3bc703f: close button is now `position: fixed`, zero-height reports floor to 1px and let the Rust min-clamp shrink to 120x80, the iframe drag relay listens in the capture phase, and `close` now carries a per-window secret token unreadable by the sandboxed iframe (with tests). Resolving the stale threads. (sent by an AI agent via Claude Code)

andrewgazelka:
Addressed the empty-card P2: `#ix-root` now has `min-width:120px; min-height:80px` matching the Rust window min-clamp, so an empty/zero-height resource renders a small visible tinted card instead of a transparent click-intercepting window. (sent by an AI agent via Claude Code)

PR #1374, opened 2026-06-19
feat(ix-vt): OscTitleTracker for the OSC window title

andrewgazelka:
## What

Add `OscTitleTracker` to `ix-vt`: a safe wrapper over ghostty's streaming OSC
parser that captures the live OSC window title (OSC 0/2) from a raw terminal
byte stream.

## Why

`ix_vt::Terminal` exposes no window-title getter (`GhosttyTerminalData` has no
title key), so a consumer that labels a session by its live title cannot read it
from a render `Snapshot`. The downstream need is the `ix` multiplexer
([ixmux](https://github.com/andrewgazelka/ixmux)) swapping its `vt100` emulator
for `ix-vt`; `tui` could use it too (it also has no title source today).

## How

A small `ESC ] … (BEL | ESC \)` framing scanner extracts OSC payloads and feeds
them to ghostty's `ghostty_osc_*` parser, so:

- sequences split across reads are still parsed (the framing state persists
  across `feed` calls, including a split `ST` terminator);
- OSC 0/2 (title) vs OSC 1 (icon) classification is done by ghostty's parser,
  not a hand-rolled scan.

The captured title is copied out of the parser-owned (transient) string into an
owned `String`.

## Tests

`tests/osc_title.rs`: BEL- and ST-terminated titles, OSC 0, a title split across
three feeds, persistence across non-title output, OSC 1 ignored, and
replacement. Validated locally (`cargo test -p ix-vt`: osc_title 4 + round_trip
3 pass) and `nix build .#ix-vt`.

> [!NOTE]
> ### Add `OscTitleTracker` to parse OSC window title sequences from terminal output
> - Introduces `OscTitleTracker` in [lib.rs](https://github.com/indexable-inc/index/pull/1374/files#diff-380b2028e1ad0ef1c65bcc54363cac0bcf356148601c2fcfe909e43b391958a0), a state machine that scans raw terminal byte streams for OSC sequences and extracts the window title.
> - Implements a `FrameState` enum to track framing across chunked feeds, handling BEL-terminated, ST-terminated (`ESC \`), and ESC-terminated sequences, with correct abort on `CAN`/`SUB`.
> - Delegates OSC payload parsing to the ghostty OSC parser via FFI (`ghostty_osc_new`, `ghostty_osc_next`, `ghostty_osc_end`, `ghostty_osc_free`).
> - Captures the title only for OSC 0/2 commands with valid UTF-8 content; leaves the previous title intact on invalid input or non-title commands.
> - Test coverage in [osc_title.rs](https://github.com/indexable-inc/index/pull/1374/files#diff-2eb4f71823e84e1e1a50a0b1bc1877727864c72c0c9b7ed5b644145c82e28133) covers BEL/ST termination, OSC 0, mid-sequence splits, invalid UTF-8, `CAN` abort, and OSC 1 exclusion.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 275c0e0.</sup>
> 
>

### DIFF
--- a/packages/tui/vt/ix-vt/src/lib.rs
+++ b/packages/tui/vt/ix-vt/src/lib.rs
@@ -701,18 +701,17 @@ const ST_FINAL: u8 = b'\\';
 const CAN: u8 = 0x18;
 const SUB: u8 = 0x1a;
 
-/// Where the framing scanner is within an `ESC ] … (BEL | ESC \)` OSC sequence.
+/// Where the framing scanner is within an `ESC ] … (BEL | ESC | CAN | SUB)` OSC
+/// sequence.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum FrameState {
     /// Outside any escape sequence.
     Ground,
-    /// Saw `ESC`; a following `]` opens an OSC.
+    /// Saw `ESC`; a following `]` opens an OSC (and a following `\` after an OSC
+    /// is the tail of a String Terminator, which lands back in `Ground`).
     Escape,
     /// Inside an OSC payload, forwarding bytes to the parser.
     Osc,
-    /// Inside an OSC payload and just saw `ESC`; a following `\` is the ST
-    /// terminator, anything else aborts the sequence.
-    OscEscape,
 }
 
 /// Tracks the most recent OSC window title in a raw terminal byte stream.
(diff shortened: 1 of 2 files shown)


andrewgazelka:
Thanks — addressed the AI-review findings in the latest commit (275c0e02):

- **Avoid materializing unknown OSC command enums (P1):** `finish` no longer calls `ghostty_osc_command_type`; it asks `ghostty_osc_command_data` for the title string and uses its bool return to classify, so no `GhosttyOscCommandType` is formed from the FFI return.
- **Finish OSC when ESC starts another sequence / Replay aborted OSC escapes:** `ESC` inside an OSC now finishes+dispatches the title and re-enters `Escape` (dropped the `OscEscape` state), matching ghostty. Added a test.
- **Don't lossy-decode malformed titles:** now `to_str()` and skip on invalid UTF-8. Added a test.
- **C0 / CAN-SUB inside OSC:** `CAN`/`SUB` abort the string; other bytes are forwarded to ghostty's own OSC parser, which applies its C0 rules.
- **C1 (8-bit) OSC framing:** declined — the agents this serves emit 7-bit OSC, and 0x9d/0x9c are ambiguous in a UTF-8 stream; documented as 7-bit-only.

PR #1375, opened 2026-06-19
nix-web-monitor: add home/os switch with live activation + diff views

andrewgazelka:
## What

Adds `nh`-style subcommands to `nix-web-monitor` (`nwm`) so a `home-manager switch` / `darwin-rebuild switch` is observable in the live browser UI, with the **activation phase** (the part you actually wait on) rendered as a live step tree:

- `nwm home switch [FLAKE] [-u]` — build `homeConfigurations."<user>@<host>".activationPackage`, then run `<out>/activate`.
- `nwm os switch [FLAKE] [-u]` — build `darwinConfigurations."<host>".system`, then activate via the canonical `sudo <out>/sw/bin/darwin-rebuild activate` (interactive `sudo -v` pre-auth first).
- `nwm flake update [INPUT...] [--flake DIR]` — update inputs under the monitor. `-u` on a switch runs the update first.
- `nwm home/os build` — build only, no activation.

Everything else still passes straight to `nix` (clap `external_subcommand`), so `nwm build .#x` and `nwm run .#ix -- new` are unchanged.

## How

A switch runs as a phase sequence under one monitor: optional flake update → monitored `nix build` (existing build tree) → activation → `nvd diff`. Activation and the diff are new **out-of-band subtrees**, mirroring the existing daemon-panel pattern (pure wire type in `parser/`, a `Delta` variant, a setter, a Svelte panel) — the build tree is reused unchanged.

- `parser/src/activation.rs` (new): pure, tested line classifier. home-manager's `Activating <step>` lines become named steps (anchored to a single-token name so a step's own output can't split it); `Starting Home Manager activation` is the banner; nix-darwin's unstructured activate streams under one seeded `activate` step.
- New `Delta::ActivationSet` / `DiffSet`, `MonitorState`/`MonitorSnapshot` `activation`/`diff` fields + setters.
- `server/src/main.rs`: subcommand restructure (`global = true` flags so they parse after a subcommand), `build_job`/`run_switch`/`run_activation`/`run_diff`, and `run_nix_build` (the old `run_nix_command` minus `finish`, with stdout capture for `--print-out-paths`).
- Svelte: `ActivationPanel.svelte` (step tree, live durations, expand-to-output) + `DiffPanel.svelte` (self-hides off a switch), wired through `types.ts`/`monitor-transport.ts`/`App.svelte`.
- `server/default.nix`: bundles `pkgs.nvd` via `--suffix PATH` (operator's own `nvd` still wins); `nix`/`home-manager`/`darwin-rebuild`/`sudo` stay on the operator's PATH as before.

Failure contract: flake-update fail → abort before build; build fail (or zero out-paths) → skip activation+diff with a distinct status and a non-zero exit; activation fail → skip diff, mark the open step failed, propagate the exit code.

## Testing

- `cargo test` for both crates green (parser classifier incl. anchored-match + darwin fallback + failed-step; `Delta::ActivationSet`/`DiffSet` msgpack round-trip).
- `svelte-check` + `eslint` + `vite build` clean.
- `nix build .#nix-web-monitor` builds.
- CLI parsing verified for every form incl. backward-compat passthrough and globals-after-subcommand.
- Reviewed by the `code-reviewer` agent (no blockers; its Low findings on the flake-update phase, the build-no-output-path edge, and the activation step-boundary match are applied in this PR).

Note: `os switch` mutates the system and needs an interactive terminal for the sudo prompt.

> [!NOTE]
> ### Add `home`/`os` switch subcommands with live activation and diff views to nix-web-monitor
> - Adds `nwm home switch` and `nwm os switch` CLI subcommands that orchestrate: optional `nix flake update`, a monitored build phase, live activation streaming, and a post-switch generation diff via `nvd`.
> - Introduces an activation model ([activation.rs](https://github.com/indexable-inc/index/pull/1375/files#diff-cf1946bcbc2504134710825012a483875144980d02aeef883e18724cbe4ae287)) that parses activation output into named steps with status, timestamps, and captured lines, broadcast as `ActivationSet` WebSocket deltas.
> - Adds `ActivationPanel` and `DiffPanel` Svelte components to the UI, showing live step progress and the post-switch diff when available.
> - Wraps the server binary with `nvd` on PATH so generation diffs work by default ([default.nix](https://github.com/indexable-inc/index/pull/1375/files#diff-b66bbd31984b70403c9f30b1a7e014fab90ee09c73a1f53dad8da7ae90823b6e)).
> - Risk: `os switch` requires `sudo` for profile registration and activation; missing `sudo` auth will stall the process.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a915399.</sup>
> 
>

### DIFF
--- a/packages/nix/nix-web-monitor/parser/src/activation.rs
+++ b/packages/nix/nix-web-monitor/parser/src/activation.rs
@@ -1,0 +1,239 @@
+//! Live activation view for `nwm home switch` / `nwm os switch`.
+//!
+//! A "switch" is two phases Nix's internal-json stream does not cover: the
+//! `nix build` of the toplevel (which the build tree already renders) and then
+//! running the generation's `activate` script. The activate script is where the
+//! work the operator actually waits on happens (linking files, installing
+//! packages, restarting launch agents), and it reports its own progress as plain
+//! stdout lines, not internal-json. This module folds those lines into a small,
+//! wire-friendly [`Activation`] subtree the UI renders beside the build tree.
+//!
+//! home-manager emits a fixed, named step sequence via `_iNote "Activating %s"`,
+//! so each `Activating <step>` line opens a step and closes the previous one,
+//! giving a real step tree. nix-darwin's `activate` is unstructured, so the
+//! server seeds a single step up front and every line lands under it.
+//!
+//! Kept pure and `now_ms`-injected (like [`crate::daemon::DaemonTrace`]) so the
+//! classifier is unit-testable without a clock; the server stamps the time and
+//! drives it through [`crate::MonitorState`].
+
+use serde::{Deserialize, Serialize};
+
+/// Prefix home-manager prints (via `_iNote "Activating %s"`) before each
+/// activation step. The remainder of the line is the step name.
+const ACTIVATING_PREFIX: &str = "Activating ";
+
+/// home-manager's banner line at the top of activation. Recognised so a stray
+/// leading line does not become an unnamed step; carries no step itself.
+const HM_BANNER: &str = "Starting Home Manager activation";
+
+/// Status of one activation step.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ActivationStatus {
+    Running,
+    Done,
+    Failed,
+}
+
+/// One activation step: a named unit of the activate script's work plus the
+/// lines it printed while running.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ActivationStep {
+    pub name: String,
+    pub status: ActivationStatus,
+    /// Output lines captured while this step was the open one.
+    pub lines: Vec<String>,
+    pub started_at_ms: u64,
+    pub stopped_at_ms: Option<u64>,
+}
+
+/// Wire-friendly snapshot of the activation phase.
+///
+/// `active` mirrors `DaemonInfo::tracing` (the panel greys out until a switch
+/// starts); `status` mirrors `DaemonInfo::status` (a human line explaining the
+/// current phase, e.g. "running", "skipped (build failed)", "failed").
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Activation {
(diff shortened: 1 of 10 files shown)


andrewgazelka:
Thanks for the review — addressed in `96a80ee2`:

**Fixed**
- **Record the nix-darwin system generation (P1)** — confirmed against the `darwin-rebuild` source: `nix-env -p <profile> --set` is gated on the `switch` action only, so `activate` alone never advances the profile. `os switch` now runs `sudo nix-env -p /nix/var/nix/profiles/system --set <out>` before activating, mirroring `darwin-rebuild switch`.
- **Pass through non-update flake subcommands** — removed the dedicated `flake` subcommand entirely; `nwm flake update …`, `nwm flake show`, `nwm flake metadata`, etc. now fall through the `external_subcommand` straight to `nix` (monitored). The `-u` switch flag keeps its own internal `nix flake update`.
- **Mark builds successful before activation can fail** — the build phase now promotes its rows to `succeeded` as soon as the build exits 0 (shared with `finish`'s promotion), so a later activation failure no longer leaves built derivations stuck as `stopped`.
- **Match nix-darwin's default configuration name** — `os` now defaults the config name via `scutil --get LocalHostName` (was `hostname -s`), matching `darwin-rebuild`.
- **Leave a result for build-only switches** — `nwm home/os build` now prints the captured store path to stdout.

**Acknowledged (kept as-is)**
- **Leading Nix options in passthrough** — with subcommands, `nwm`'s own flags precede the command and Nix flags follow the Nix subcommand (`nwm build .#x --option …`, which works). Leading global Nix flags before the subcommand are the one ordering that changed; the common form is preserved.
- **Activity ids across update + build under `--update`** — the update phase's activities complete before the build phase starts, so any id reuse only affects the ephemeral update rows, not the build tree. Left as a known minor cosmetic limitation of the opt-in `-u` path.

🤖 via Claude Code

andrewgazelka:
Second-round review (ran against `274c5709`, before the fixes landed):
- **os switch doesn't record the generation (P1)** and **flake passthrough (P2)** — already fixed in `96a80ee2` (set `/nix/var/nix/profiles/system` before activate; removed the `flake` subcommand so `nix flake show/check/...` pass through).
- **Activation output ignores `--terminal-output quiet` (P3)** — fixed in `a9153992`: `TerminalOutput` is threaded into `stream_activation`, which suppresses the per-line echo under `Quiet` (the browser still gets every line).

🤖 via Claude Code

Issue #1415, opened 2026-06-20
blast-radius PR comment "comes up empty": comment job silently skipped when eval bails


andrewgazelka:
## Symptom

A teammate reported the **blast-radius PR comment "sometimes comes up empty"** on index PRs.

## Findings (from live GitHub state, not memory)

1. **Currently disabled on PRs.** #1384 (merged 2026-06-19) swapped the `pull_request` trigger for `workflow_dispatch` to save shared `ix-ci` runner time. So today the workflow posts **no** comment on any PR — that is the immediate "empty".

2. **Historical "empty" = the `comment` job being silently skipped.** Across the last ~20 failed `Blast radius` runs, *every* failure was in the `evaluate` job's **"Report blast radius"** step (`nix run .#blast-radius` exiting non-zero), after which the `comment` job (`needs: evaluate`, default `if: success()`) is **skipped** — so no sticky comment is posted or updated and the PR shows nothing.

3. **The bail is often base-side, not the PR's fault.** Examples: a docs-only RFC PR (`dev-vm-environments`, runs 27574479677/27574011741) failed because the **base** (`master`) catalog was transiently un-evaluable (`tests // { recurseForDerivations = ...; }` in `lib/per-system.nix` where `tests` was not an attrset). When master is red / a runner flakes, blast-radius can't diff and the comment silently vanishes on otherwise-fine PRs.

4. The renderer always emits at least a header, so a *posted* comment is never literally blank. "Empty" means **absent/skipped**, not blank.

## Fix

Make the `comment` job resilient so a PR never ends up with a silently-absent blast-radius comment:
- Run the `comment` job even when `evaluate` fails (`if: !cancelled() && needs.evaluate.result != 'skipped'`).
- When `report.json` is absent (eval failed), post an explicit sticky note ("could not compute the blast radius for this run; usually a transient base/runner issue, see logs") instead of skipping. A later good run overwrites it in place (same `<!-- blast-radius -->` marker).
- Keep the eval fail-closed semantics and the (deliberately disabled, per #1384) trigger untouched. Re-enabling the per-PR trigger is a separate call for the team.

sent by an AI agent (Claude) via Claude Code

andrewgazelka:
Closing: auto-filed by a runaway system-prompt eval rollout (live `claude -p` agents that ran tasks end-to-end), not a real request. Root cause fixed in #1471. (cleanup by an AI agent via Claude Code)

Issue #1411, opened 2026-06-20
blast-radius PR comment "comes up empty" = absent (disabled in #1384); make it opt-in via a label


andrewgazelka:
**Symptom (teammate report):** the blast-radius PR comment "sometimes comes up empty."

**Investigation (evidence):**
- Comment *body* is never empty: 368/368 blast-radius comments repo-wide have full content; 0 report `0 of N`, 0 miss the rebuild line.
- Mermaid never breaks: all 49 distinct flowcharts render cleanly in mermaid-cli 11.12 (headless Chrome); causes are capped (`max_causes=6`), so the diagram is always tiny.
- The real symptom is the comment being **absent**: of the last 200 runs, ~30% posted nothing — 18 `cancelled` (cancel-in-progress on a new push) + 41 `action_required` (fork/approval-gated).
- As of #1384 (~2026-06-19) the `pull_request` trigger was commented out entirely for shared `ix-ci` runner-slot reasons, so new PRs now get no comment at all.

**Fix:** make it opt-in via a `blast-radius` PR label rather than all-or-nothing. Restores a reliable on-demand signal without queuing an eval-only runner on every PR (respects #1384). Label-add + each later push refreshes the sticky comment; unlabeled PRs skip `evaluate` and claim no runner.

andrewgazelka:
Closing: auto-filed by a runaway system-prompt eval rollout (live `claude -p` agents that ran tasks end-to-end), not a real request. Root cause fixed in #1471. (cleanup by an AI agent via Claude Code)

PR #1424, opened 2026-06-20
ci(blast-radius): opt-in label trigger + resilient comment job

andrewgazelka:
## Problem

A teammate reported the **blast-radius PR comment "sometimes comes up empty"** on index PRs. This consolidates the two open fixes (#1413, #1416) into one coherent change.

## Root cause (live evidence, not memory)

Pulled the **last 200 `Blast radius` runs**: `0 failure`, 141 success, **41 `action_required`** (100% `update-flake-lock`/`update-cargo` bot PRs), **17 `cancelled`** (cancel-in-progress on rapid pushes). And **368/368** posted comment bodies have full content.

So the body is **never blank** — "empty" means the comment is **absent**:
- **#1384** (merged earlier today) commented out the `pull_request` trigger entirely to save a shared `ix-ci` runner slot, so **every PR now gets no comment** — the dominant current cause.
- `action_required` (bot PRs) and `cancelled` (self-healing) post nothing; and an `evaluate`-job failure (a base-side eval bail or runner flake) **skips** the `comment` job (`needs: evaluate`, default `if: success()`), leaving a labeled PR silently commentless.

## Fix

Matches the **#general "should we make blast radius opt in?" thread** (two 👍 incl. "+1 opt in"):

1. **Opt-in trigger** (from #1413): re-enable `pull_request` for `types: [labeled, synchronize, reopened]`, and gate the `evaluate` job's `if` on the `blast-radius` label. Unlabeled PRs skip `evaluate` and **claim no runner** (respects #1384); once a maintainer adds the label, each later push refreshes the sticky comment.
2. **Resilient comment job** (from #1416): run `comment` even when `evaluate` fails and post an explicit **"could not compute"** sticky note (same `<!-- blast-radius -->` marker, overwritten in place by the next good run) instead of skipping. Every gated step restates `success()` so the validate → render → post chain stays **fail-closed** (an explicit `if:` drops GitHub's implicit `success()`).

## Test

`tools/blast-radius-test.sh` extracts the workflow's jq/scripts **verbatim** so the test can't drift from the trusted job. Added the fallback-render case and fail-closed gating assertions. Full suite passes (`jq` + `yq-go`); `actionlint` clean.

Supersedes #1413 and #1416.

Fixes #1415
Fixes #1411

sent by an AI agent (Claude) via Claude Code

> [!NOTE]
> ### Add opt-in label trigger and fallback comment to blast-radius workflow
[...]

Issue #1421, opened 2026-06-20
blast-radius PR comment sometimes empty: validator rejects nix names with `?`/`=`


andrewgazelka:
## Symptom
The blast-radius sticky PR comment **sometimes comes up empty** (no comment posts at all) on index PRs.

## Root cause
The trusted `comment` job in `.github/workflows/blast-radius.yml` validates every check/cause/timing name against:

```
name_ok: ^[A-Za-z0-9_./ +():-]+$
```

and **fail-closes the entire report** (`::error::malformed`, `exit 1`) if *any* single name falls outside that charset. But **cause names are nix derivation names**, which legally contain `?` and `=` (store-path-name charset is `[a-zA-Z0-9+._?=-]`). Fixed-output patch/fetch derivations surface as frontier causes with names like:

- `e67caa0...patch?full_index=1`
- `webkitgtk-2.52.4+abi=4.1`

Whenever a PR's rebuild frontier includes such a derivation, the whole report is rejected and **no comment posts**. The Rust producer's local `to_markdown` has no such guard, so it only reproduces in CI and only "sometimes" (data-dependent on the rebuilt set).

## Repro (MRE)
Feed the trusted validator a schema-valid report whose cause name is a real patch derivation:
```json
{"base":"aaaaaaa","head":"bbbbbbb","total":120,"changed":["rust-test-foo"],"added":[],"removed":[],
 "causes":[{"name":"e67caa006c75181b45b761cd50294cb3c8e18f1a.patch?full_index=1","checks":["rust-test-foo"]}],
 "timings":{},"phaseTimings":{"total":1.0}}
```
→ `::error::malformed blast-radius report.json`, job exits 1, no comment.

28+ such names exist in a local `/nix/store` sample.

## Fix
Widen `name_ok` / `safename` (kept in lockstep) to the actual legal nix-name set by adding `?` and `=`:
```
^[A-Za-z0-9_./ +()?=:-]+$
```
Verified to still reject every markdown/mention/HTML injection char (`@ < > [ ] ` `` ` `` & | ; * # {`). Add a `?`/`=` fixture to `tools/blast-radius-test.sh`.

Found while investigating; PR incoming.

andrewgazelka:
**The fix is correct and validated — the blocker is a runaway agent swarm, not the code.**

[...]

PR #1425, opened 2026-06-20
ci(blast-radius): post a fallback comment when the report cannot be rendered

andrewgazelka:
## Symptom

A teammate reported the **blast-radius PR comment "sometimes comes up empty"** on index PRs.

## Root cause (from live state, not memory)

- **The renderer never posts a blank body.** Reproduced the workflow's own validate+render jq against good / zero-change / overflow / pathological reports: it always emits at least the `### Blast radius` header. Confirmed in the wild too (prior scans found 368/368 blast-radius comments well-formed). So **"empty" means the comment is absent.**
- **The real bug: the `comment` job is `needs: evaluate` with the default `if: success()`.** When `evaluate` bails, `comment` is **silently skipped** and the PR is left with no comment. The eval bail is frequently **base-side** (a transiently un-evaluable `main`) or a runner flake, i.e. nothing to do with the PR. Across recent failed `Blast radius` runs, every failure was in `evaluate`'s "Report blast radius" step.
- (Separately, #1384 commented out the `pull_request` trigger entirely for shared-runner cost. Re-enabling it is a team call, tracked in #1411/#1413 and **left untouched here**.)

## Fix

Make the `comment` job resilient so a PR never silently ends up without a blast-radius comment:

- Run `comment` even when `evaluate` failed: `if: ${{ !cancelled() && needs.evaluate.result != 'skipped' }}` (still skipped on forks/Dependabot, where `evaluate` itself was skipped).
- Gate `Download report` / `Validate` / `Render` on `evaluate` success; `Download` is `continue-on-error` so a transient artifact miss falls through to the fallback.
- New **Compose fallback comment** step fires whenever `render` did not succeed (eval failure, missing artifact, *or* a render error), writing an explicit "could not compute" note under the same `<!-- blast-radius -->` marker so the next good run overwrites it in place.
- **Post only a producer that SUCCEEDED** (`steps.render.conclusion == 'success' || steps.fallback.conclusion == 'success'`), so a half-written `comment.md` from a failed render is never patched onto the PR.

This last point is the correctness fix the AI-review gate flagged on **#1416** (its `Post` step ran on bare `!cancelled()` and could publish a partial body). This PR **supersedes #1416** — same approach, with that gap closed — so #1416 and the duplicate investigations can be closed in favor of this one.

## Test

Extended `tools/blast-radius-test.sh` (which extracts the workflow's own step scripts, so it can't drift) with a fallback case: runs the extracted fallback script with no report and asserts a marker-prefixed body + the "could not compute" note + the run link.

Validated locally:
- `tools/blast-radius-test.sh` → all passed (incl. new fallback case)
- `actionlint .github/workflows/blast-radius.yml` → clean
- adversarial code-review of the diff (named subagent) → only the post-gating concern, fixed here

Fixes #1415

> [!NOTE]
> ### Post a fallback comment in the blast-radius workflow when report rendering fails
> - The `comment` job in [blast-radius.yml](.github/workflows/blast-radius.yml) now runs when `evaluate` fails (but not when skipped), so failures are always surfaced on the PR.
> - Download and render steps run only when `evaluate` succeeded; download errors are tolerated via `continue-on-error`.
> - A new 'Compose fallback comment' step runs when rendering did not succeed and posts a minimal marker-prefixed comment with a link to the Actions run URL.
> - The 'Post sticky comment' step only fires when either the rendered or fallback body is available, preventing partial posts.
> - A new test block in [blast-radius-test.sh](https://github.com/indexable-inc/index/pull/1425/files#diff-008b7e349202f69865582328028ef88fb2c6c22cdfd1f42fbbf5533600c2e27b) validates the fallback comment output.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f70b473.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -125,6 +125,15 @@ jobs:
   # PR cannot influence.
   comment:
     needs: evaluate
+    # Run even when `evaluate` failed, so a PR never silently ends up with no
+    # blast-radius comment (the "comes up empty" report, #1415). The eval bail
+    # is frequently base-side (a transiently un-evaluable `main`) or a runner
+    # flake, i.e. unrelated to this PR. `!cancelled()` covers success and
+    # failure but not a concurrency cancel (which cancels this job too, so its
+    # own `cancelled()` is true); `result != 'skipped'` keeps it off forks and
+    # Dependabot, where `evaluate` itself was skipped and there is nothing to
+    # report.
+    if: ${{ !cancelled() && needs.evaluate.result != 'skipped' }}
     runs-on: ubuntu-latest
     permissions:
       contents: read
@@ -134,7 +143,13 @@ jobs:
       REPOSITORY: ${{ github.repository }}
       GH_TOKEN: ${{ github.token }}
     steps:
+      # Present only when `evaluate` succeeded (its upload step is `if: success()`).
+      # `continue-on-error` so a transient download failure does not fail the job
+      # outright: the render step then no-ops (no report.json) and the fallback
+      # composer (gated on render not having succeeded) takes over.
       - name: Download report
+        if: ${{ needs.evaluate.result == 'success' }}
+        continue-on-error: true
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: blast-radius-${{ github.run_id }}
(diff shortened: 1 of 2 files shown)


andrewgazelka:
Superseded by #1424, which is the consolidated fix. #1424 keeps your resilient eval-failure fallback but, unlike this PR, leaves a *render* failure fail-closed (a red check) rather than posting the "failed to evaluate at base/head" note — that message is inaccurate when eval actually succeeded and only the deterministic jq render broke. #1424 also re-enables the trigger (opt-in via the `blast-radius` label), without which the comment never runs at all on current PRs. Closing in favor of #1424 to keep one implementation.

andrewgazelka:
Superseded by #1428 (single consolidated fix for #1415 + #1421). sent by an AI agent (Claude) via Claude Code

PR #1426, opened 2026-06-20
fix(blast-radius): stop the PR comment coming up empty (validator charset + opt-in re-enable)

andrewgazelka:
## What
The blast-radius sticky PR comment **comes up empty** on index PRs. This fixes it end to end and restores the comment. Fixes #1421.

## Two coupled causes
1. **The bug (why it was empty).** The trusted `comment` job validates every check/cause/timing name against `name_ok = ^[A-Za-z0-9_./ +():-]+$` and **fail-closes the whole report** (`exit 1`) if any name is outside it. But **cause names are nix derivation names**, whose legal charset includes `?` and `=`. Fixed-output patch/fetch drvs surface as frontier causes named like `<sha>.patch?full_index=1` and `webkitgtk-2.52.4+abi=4.1` (86 such names in a local `/nix/store` sample). Any PR whose rebuild frontier included one → whole report rejected → **no comment**. The Rust producer's local `to_markdown` has no such guard, so it only reproduced in CI and only "sometimes".
2. **Why it wouldn't appear at all.** #1384 disabled the per-PR run entirely (shared `ix-ci` runner cost on every PR). So fixing the charset alone leaves the comment never showing. The two must land together.

## Fix
- **Charset (root cause):** widen `name_ok` (validator) and `safename` (renderer) **in lockstep** to `^[A-Za-z0-9_./ +()?=:-]+$` — the legal nix-name glyphs that reach a label. Still rejects every markdown/mention/HTML glyph (`@ < > [ ] \` & | ; * # {`, newline). Update the `nix.rs` doc comment.
- **Opt-in re-enable (so it actually appears, without #1384's queue cost):** trigger on `pull_request` `labeled`/`synchronize`/`reopened`, gated by a `blast-radius` label on the `evaluate` job. Unlabeled PRs claim no `ix-ci` runner (jobs skip), so the queue concern #1384 fixed stays fixed; add the label to opt a PR in.
- **Resilient comment:** the `comment` job now runs on eval success **or** failure (skips only when `evaluate` is skipped). On failure it posts a "could not compute (transient)" note with the run link under the same sticky marker, instead of leaving the PR silent.
- **Tests:** `tools/blast-radius-fixtures/special-name.json` (cause name with `?`/`=`) + assertions, and fallback-path assertions, in `tools/blast-radius-test.sh` (the `blast-radius-test` nix check).

## Validation
- `tools/blast-radius-test.sh`: **all passed** (validate/render `special-name`, injection-rejection `bad-name`/`bad-check`, overflow/backstop, fallback marker/explanation/link).
- `actionlint .github/workflows/blast-radius.yml`: clean.
- Independently confirmed `?`/`=` appear in 86 real `/nix/store` derivation names.

Consolidates the blast-radius PR swarm: supersedes #1427 (opt-in re-enable + fallback), #1424 closed as dup.

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -153,7 +153,14 @@ jobs:
           # otherwise exits on the last value, letting `{}` followed by a valid
           # report pass while the render then emits a null first section.
           jq -e -s '
-            def name_ok: type == "string" and test("^[A-Za-z0-9_./ +():-]+$");
+            # Charset is the legal nix store-path-name set (a-zA-Z0-9 + - . _ ? =)
+            # plus the extra glyphs attr-path normalization can introduce
+            # (/, space, (, ), :). A cause name is a derivation name, and fixed-
+            # output patch/fetch drvs carry `?` and `=` (e.g. `<sha>.patch?full_index=1`,
+            # `webkitgtk-2.52.4+abi=4.1`); omitting them fail-closed the whole
+            # report and posted no comment at all (the "sometimes empty" bug, #1421).
+            # It still rejects every markdown/mention/HTML glyph (@ < > [ ] ` & | ; * # {).
+            def name_ok: type == "string" and test("^[A-Za-z0-9_./ +()?=:-]+$");
             length == 1 and
             (.[0] |
               (.base    | type == "string" and test("^[0-9a-f]{7,40}$")) and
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Superseded by #1428, which consolidates the charset root-cause fix (#1421) and the resilient/opt-in comment job (#1415) into a single PR. Closing to avoid five competing PRs for one fix.

andrewgazelka:
Consolidated the blast-radius "empty comment" fixes into #1431 (opt-in label trigger + resilient comment job + legal-name `?`/`=` charset), which closes #1411/#1415/#1421 and passes the full `blast-radius-test` suite. This overlaps with the work here, so suggest closing this in favor of #1431 once that lands (no rush, just flagging to avoid double-merge).

PR #1427, opened 2026-06-20
ci(blast-radius): restore the PR comment (opt-in label, resilient fallback, accept ?/= in nix names)

andrewgazelka:
## Problem

A teammate reported the blast-radius PR comment "sometimes comes up empty." I investigated end to end (evidence, not memory):

- **The comment body is never blank.** A repo-wide scan of every `github-actions[bot]` comment starting with `<!-- blast-radius -->` found **359/359 with full content** — 0 blank, 0 header-only, 0 `0 of N`; the lowest ever was `2 of 976`. A forced **zero-rebuild** render still produces a clear 118-byte `0 of N checks would rebuild` body. The renderer always emits marker + header + rebuild line.
- **"Empty" means the comment was *absent*.** And as of #1384 the `pull_request` trigger is commented out entirely, so **new PRs get no comment at all** — the dominant cause today.

### Root causes of a *missing* comment (both fixed here)
1. **Trigger disabled (#1384).** Commented out for shared `ix-ci` runner-slot cost.
2. **Eval bail silently skips the comment.** `comment` is `needs: evaluate` / `if: success()`; when `evaluate`'s "Report blast radius" step exits non-zero — **frequently base-side** (a transiently un-evaluable `master`) or a runner flake, i.e. unrelated to the PR — `comment` is skipped, leaving no comment.

## Fix (consolidates the two in-flight partials)

This supersedes #1413 (opt-in trigger) and #1416 (resilient fallback), which each fix only one cause and both edit this same file (so they conflict). One complete change:

- **Re-enable on PRs, opt-in via a `blast-radius` label.** Respects #1384's cost reason: unlabeled PRs claim no runner. Triggers on `labeled` / `synchronize` / `reopened`; `evaluate` is gated on `contains(...labels..., 'blast-radius')`.
- **Make `comment` resilient.** It now runs on eval success **or** failure (`if: !cancelled() && needs.evaluate.result != 'skipped'`); skipped only when `evaluate` itself was skipped (unlabeled / fork / Dependabot, which opted out). Download + validate + render stay gated on eval success; on failure a new step posts an explicit **"could not compute"** sticky note (same `<!-- blast-radius -->` marker, overwritten in place by the next good run) instead of nothing.

Eval **fail-closed** semantics (`guard_eval_failures`) are untouched — relaxing them (the deeper root of base-side flakes) is a separate design call.

## Test

`tools/blast-radius-test.sh` already extracts the workflow's jq verbatim so the test can't drift from what runs. Extended with a fallback-path case (extracts the new step, runs it, asserts the marker survives + the explanation and run link are present). Verified red→green (fails on the pre-change workflow, passes after). Full suite green; `actionlint` clean. Wired as the `blast-radius-test` flake check.

Supersedes #1413, #1416. Closes #1411, closes #1415.

> [!NOTE]
> ### Restore blast-radius PR comment with opt-in label gate and fallback on eval failure
> - The [blast-radius workflow](https://github.com/indexable-inc/index/pull/1427/files#diff-f18ec732b17d456a8d2f9c00a4941f9433314b97b8a3df9cea22bdaf15337468) now auto-triggers on `labeled`, `synchronize`, and `reopened` PR events targeting `main`, replacing the previously commented-out trigger.
> - The `evaluate` job is gated by a `blast-radius` label on the PR, in addition to same-repo and non-Dependabot checks.
> - The `comment` job now runs when `evaluate` completes (success or failure, but not skipped); on failure it posts a sticky fallback comment with an explanation and a link to the run instead of leaving no comment.
> - A new fallback-path test in [blast-radius-test.sh](https://github.com/indexable-inc/index/pull/1427/files#diff-008b7e349202f69865582328028ef88fb2c6c22cdfd1f42fbbf5533600c2e27b) validates the fallback comment contents (marker, explanation text, and run link).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0e6464e.</sup>
> 
>

---

## Update: consolidated #1426 (charset fail-close)

After independently verifying the root cause (368/368 historical bot comments had full bodies, min 183 chars — so "empty" means **absent**, not blank), this PR now also folds in the charset fix from #1426. That bug is real and complementary: a cause name is a nix derivation name, and fixed-output patch/fetch drvs legally carry `?`/`=` (86 such names in a local `/nix/store` sample, e.g. `<sha>.patch?full_index=1`). The trusted `comment` job's `name_ok` validator rejected those → fail-closed the whole report → **absent comment**. That validate step lives in the `comment` job, **not** `evaluate`, so this PR's eval-failure fallback would *not* have caught it — both fixes are required and neither alone is complete.

Widened `name_ok` + renderer `safename` in lockstep to `^[A-Za-z0-9_./ +()?=:-]+$` (still rejects every markdown/mention/HTML glyph), plus a regression fixture + assertions. Full `tools/blast-radius-test.sh` suite green (18/18, incl. special-name and fallback cases); `actionlint` clean.

Supersedes #1426.

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,17 +1,21 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in on PRs via a `blast-radius` label, not every PR. The per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claims a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, so running
+  # it on every PR slowed the queue -- why #1384 disabled the trigger entirely.
+  # Gating on the label (see the `evaluate` job `if:`) keeps that cost off
+  # unlabeled PRs while restoring the comment for PRs a maintainer opts in.
+  # `labeled` fires the first run when the label is added; `synchronize` /
+  # `reopened` refresh the sticky comment on later pushes (each re-checks the
+  # label, so removing it stops future runs). `workflow_dispatch` keeps `on:`
+  # non-empty for manual runs (PR-context gated, so a bare dispatch no-ops).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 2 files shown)


andrewgazelka:
Superseded by #1428, which consolidates the charset root-cause fix (#1421) and the resilient/opt-in comment job (#1415) into a single PR. Closing to avoid five competing PRs for one fix.

andrewgazelka:
Superseded by #1426, which now consolidates the full end-to-end fix: the validator charset root cause (empty comment) **plus** this PR's opt-in `blast-radius` label re-enable and resilient fallback comment. They must land together (re-enabling without the charset fix just resumes empty comments), so I folded this into #1426. Closing.

andrewgazelka:
Consolidated the blast-radius "empty comment" fixes into #1431 (opt-in label trigger + resilient comment job + legal-name `?`/`=` charset), which closes #1411/#1415/#1421 and passes the full `blast-radius-test` suite. This overlaps with the work here, so suggest closing this in favor of #1431 once that lands (no rush, just flagging to avoid double-merge).

PR #1430, opened 2026-06-20
fix(blast-radius): stop the PR comment coming up empty (opt-in trigger + resilient job + ?/= charset)

andrewgazelka:
## What

The blast-radius PR comment was coming up empty/missing for two distinct reasons. This consolidates the fixes for both into one change (superseding #1424, #1426, #1427) so the comment is reliable end to end.

### Root cause 1 — disabled on PRs (#1415)
After #1384 the workflow was fully disabled on `pull_request`, so the sticky comment never appeared. Re-enabled as **opt-in**: it runs only when a maintainer adds the `blast-radius` label (the per-PR full-catalog eval claims a shared `ix-ci` runner, so we don't want it on every push). The `comment` job is now resilient — when `evaluate` fails (usually a transient base-branch eval or runner flake), it posts an explicit "could not compute" fallback instead of silently skipping, keyed by the same `<!-- blast-radius -->` marker so the next good run overwrites it in place. The produce→post chain stays **fail-closed**: every gated step re-states `success()`, so a malformed report or half-written `comment.md` is never published.

### Root cause 2 — validator rejected legal nix names (#1421, the "sometimes empty")
The schema `name_ok` / render `safename` charset `^[A-Za-z0-9_./ +():-]+$` omitted `?` and `=`. A cause name is a derivation name, and fixed-output patch/fetch drvs legally carry both (e.g. `<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). Whenever a PR's blast radius touched such a drv, validation fail-closed the **whole** report → render + post skipped → no comment. That's the intermittent symptom. Charset extended to the legal nix store-path-name set `^[A-Za-z0-9_./ +()?=:-]+$` in both the validator and the renderer (kept lockstep); it still rejects every markdown/mention/HTML glyph.

## Evidence (MRE)
The current validator exits `1` on a real fetchpatch cause name `...patch?full_index=1`; the fixed charset accepts it and the `abi=4.1` form while still rejecting `@evil`, backticks, and `[x](url)`.

## Tests
`tools/blast-radius-test.sh` (wired as the `blast-radius-test` flake check) gains: a `special-name.json` fixture asserting the report both validates and renders with the `?`/`=` name intact; fallback-comment assertions; and fail-closed gating-wiring assertions. All 20 assertions pass locally.

Closes #1415
Closes #1421

> [!NOTE]
> ### Fix blast-radius PR comment being empty by adding opt-in label gate and fallback comment
> - The [blast-radius workflow](.github/workflows/blast-radius.yml) now triggers on `pull_request` events (labeled, synchronize, reopened) but only runs the evaluate job when the PR has the `blast-radius` label, skipping Dependabot and fork PRs.
> - When evaluation fails, a fallback comment explaining the failure is posted instead of leaving no comment.
> - The jq schema validator and renderer now accept `?` and `=` in cause names, fixing validation failures for valid nix derivation names. A [test fixture](tools/blast-radius-fixtures/special-name.json) and tests in [blast-radius-test.sh](tools/blast-radius-test.sh) cover this and the fallback behavior.
> - Behavioral Change: the evaluate job is now skipped by default unless a maintainer applies the `blast-radius` label.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 551b441.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,17 +1,22 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in per PR, not on every push. Auto-running on every PR meant the per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claimed a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, queuing an
+  # eval-only runner each push and slowing the queue (#1384). But fully disabling
+  # it left the report unreachable, so the sticky comment "came up empty" on every
+  # PR. The middle ground: it runs only when a maintainer adds the `blast-radius`
+  # label. Once labeled, each later push (`synchronize`) and reopen refreshes the
+  # comment. The job `if:` below gates EVERY event on the label, so adding any
+  # other label (or pushing to an unlabeled PR) skips `evaluate` and claims no
+  # runner. `workflow_dispatch` is kept for manual reruns (no-ops without PR
+  # context).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Closing as a duplicate — consolidated into #1428 (identical/superset diff: charset widen + resilient comment job + opt-in label trigger). #1428 is the canonical PR being driven to merge.

andrewgazelka:
Reopening as the single canonical fix. A swarm of parallel agent runs on this task created a mutual-supersession deadlock (#1428 ↔ #1430 each closed citing the other), so nothing ever merged. Stopping that here.

Independently reproduced and validated before reopening:
- **Root cause (concrete bug, #1421):** the trusted comment job's `name_ok`/`safename` regex `^[A-Za-z0-9_./ +():-]+$` rejects legal nix derivation names containing `?`/`=` (fixed-output patch/fetch drvs like `<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). A frontier cause with such a name fail-closes the whole report → `exit 1` → comment job fails → **no comment**. "Sometimes" = data-dependent on the rebuild frontier.
- **Premise check:** the renderer never emits a blank body (a degenerate `total=0` report still renders a header), so "empty" = **absent**. Reinforced by #1384 disabling the PR trigger and eval-bail skips (#1415).
- **Verified:** all 20 `tools/blast-radius-test.sh` assertions pass on this branch (`nix shell nixpkgs#yq-go nixpkgs#jq -c bash tools/blast-radius-test.sh`); the widened charset still rejects every markdown/mention/HTML injection glyph.

Driving this one to merge via the queue.

PR #1432, opened 2026-06-20
fix(blast-radius): accept nix names with ? and = so the PR comment stops coming up empty (#1421)

andrewgazelka:
## What

The trusted `comment` job in `.github/workflows/blast-radius.yml` validates every check/cause/timing name against `name_ok = ^[A-Za-z0-9_./ +():-]+$` and **fail-closes the WHOLE report** (`exit 1`, no comment) if any one name falls outside it.

But cause names are nix derivation names, whose legal charset includes `?` and `=` (e.g. `<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). Whenever a PR's rebuild frontier includes such a fixed-output patch/fetch derivation, the comment job aborted and posted **nothing** — the "sometimes empty" symptom (#1421). The Rust producer's local `to_markdown` has no such guard, so it only reproduced in CI and only on data-dependent PRs (dependency bumps, toolchain changes).

## Fix

Widen `name_ok` and the renderer's `safename` (kept lockstep) to `^[A-Za-z0-9_./ +()?=:-]+$` — the legal nix-name glyphs that reach a label. Still rejects every markdown/mention/HTML glyph (`@ < > [ ] \` & | ; * # {`). Adds a regression fixture + assertions to `tools/blast-radius-test.sh`.

## Validation

`bash tools/blast-radius-test.sh` passes locally (yq-go + jq):
- `validate special-name: ok` / `render special-name: name preserved ok` (the #1421 regression)
- `validate bad-name: rejected ok` (injection still fail-closed)
- all existing schema/overflow/backstop assertions green

Closes #1421.

> Note: blast-radius is currently disabled on PRs (#1384) so this has no visible effect until the workflow is re-enabled (see #1427's opt-in work). This lands the root-cause correctness fix so the comment can't come up empty once re-enabled.

> [!NOTE]
> ### Fix blast-radius PR comment rendering to accept `?` and `=` in nix derivation names
> - Updates the `name_ok` jq regex in [blast-radius.yml](.github/workflows/blast-radius.yml) to allow `?` and `=` characters, which are valid nix derivation name glyphs, so the schema validator no longer rejects entire reports containing them.
> - Updates the `safename` jq regex in the render step to match, preserving labels with `?` and `=` instead of dropping them and producing an empty PR comment.
> - Adds a [special-name.json](tools/blast-radius-fixtures/special-name.json) fixture and a regression test in [blast-radius-test.sh](tools/blast-radius-test.sh) to catch regressions for names containing these characters.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fbd506d.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -153,7 +153,14 @@ jobs:
           # otherwise exits on the last value, letting `{}` followed by a valid
           # report pass while the render then emits a null first section.
           jq -e -s '
-            def name_ok: type == "string" and test("^[A-Za-z0-9_./ +():-]+$");
+            # Charset is the legal nix store-path-name set (a-zA-Z0-9 + - . _ ? =)
+            # plus the extra glyphs attr-path normalization can introduce
+            # (/, space, (, ), :). A cause name is a derivation name, and fixed-
+            # output patch/fetch drvs carry `?` and `=` (e.g. `<sha>.patch?full_index=1`,
+            # `webkitgtk-2.52.4+abi=4.1`); omitting them fail-closed the whole
+            # report and posted no comment at all (the "sometimes empty" bug, #1421).
+            # It still rejects every markdown/mention/HTML glyph (@ < > [ ] ` & | ; * # {).
+            def name_ok: type == "string" and test("^[A-Za-z0-9_./ +()?=:-]+$");
             length == 1 and
             (.[0] |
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Closing as a duplicate — consolidated into #1428 (identical/superset diff: charset widen + resilient comment job + opt-in label trigger). #1428 is the canonical PR being driven to merge.

PR #1437, opened 2026-06-20
ci(blast-radius): fix the empty PR comment (opt-in re-enable + per-label gate + charset + resilience)

andrewgazelka:
## Why
A teammate reported the blast-radius PR comment **"sometimes comes up empty"** on index PRs.

## Root cause (live evidence, not memory)
Pulled the last 100 `Blast radius` runs: **60 success, 24 `action_required` (all bot PRs: update-flake-lock/update-cargo), 14 `cancelled`, 0 `failure`.** A *posted* body is never blank (the renderer always emits a header) — **"empty" means the comment is ABSENT.** Reasons, dominant first:

1. **#1384 (merged earlier today) commented out the `pull_request` trigger entirely** to save a shared `ix-ci` runner slot, so every PR now gets **no** comment.
2. **Latent fail-closed bug (#1421):** the trusted `comment` job validates every cause name against a charset missing `?`/`=`. But a cause name is a *nix derivation name*, and fixed-output patch/fetch drvs legally carry those (`<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). A rebuild frontier hitting one exits the validator `1` → comment job fails → no comment. (Not yet observed in the 100-run window, so this is a latent hardening, not the dominant cause — but it is a real fail-closed path.)
3. **Eval bail skips the comment** (`needs: evaluate`, default `if: success()`) — frequently base-side (a transiently un-evaluable base) or a runner flake unrelated to the PR.

## Fix (one workflow change + lockstep renderer)
- **Opt-in, reachable again:** re-enable `pull_request: [labeled, synchronize, reopened]` gated on a `blast-radius` label, so the comment returns **without** re-introducing the #1384 queue cost. The gate matches the label **per event** (`action != 'labeled' || label.name == 'blast-radius'`) so adding an *unrelated* label to an already opted-in PR does **not** re-claim an eval runner. *(This closes the AI-review P2 that the earlier consolidation PRs carried.)*
- **Charset:** widen `name_ok`/`safename` (kept lockstep) to the legal nix-name set `^[A-Za-z0-9_./ +()?=:-]+$`; still rejects every markdown/mention/HTML glyph (`@ < > [ ] ` `` ` `` ` & | ; * # {`). Rust `normalize_attr` doc updated; new `special-name.json` fixture.
- **Resilience:** `comment` runs on eval success **or** failure and posts an explicit "could not compute" sticky note (same `<!-- blast-radius -->` marker, overwritten by the next good run) instead of skipping. Every gated step re-states `success()` so the validate→render fail-closed chain still holds.

## Tests
`tools/blast-radius-test.sh` (the `blast-radius-test` CI check; it extracts the workflow's own jq/scripts so it can't drift) now also asserts the charset regression, the fallback path, the fail-closed gate wiring, **and the per-label opt-in gate**. All 22 assertions pass locally (`yq-go` + `jq`); `actionlint` clean.

## Note on duplicates
This task spawned a cluster of overlapping PRs (#1413, #1416, #1422, #1424, #1425, #1426, #1427 closed; #1428 open). This is the consolidated, complete version: #1428's content **plus** the per-label P2 fix and its regression test. Recommend landing this one and closing the rest.

Closes #1415
Closes #1421

> [!NOTE]
> ### Fix empty PR comments in blast-radius workflow with label gating and fallback handling
> - Adds a `blast-radius` label gate to the evaluate job so the workflow only runs on labeled PRs and does not claim a runner for unrelated label additions.
> - Tightens step chaining with explicit `success()` guards on Validate and Render steps to prevent silent failures from producing empty comments.
> - Adds a fallback comment step that posts a minimal 'could not compute' note with a run link when evaluation fails but is not skipped.
> - Extends the `safename` regex and jq validator to accept `?` and `=` in derivation names, keeping validation and rendering in lockstep.
> - Behavioral Change: unlabeled PRs no longer trigger the workflow at all; previously they would run and could post an empty comment.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7b39479.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -47,17 +47,21 @@ jobs:
   # data) for the comment job.
   evaluate:
     # Same-repository, non-Dependabot PRs only (mirroring ai-review-gate.yml),
-    # and only when the PR carries the `blast-radius` label (the opt-in gate;
-    # see `on:` above). `labeled` events already carry the new label in
-    # `pull_request.labels`, so `contains` covers both the initial label-add and
-    # every later `synchronize`/`reopened` while the label remains. The `comment`
-    # job has no `if:` of its own but `needs: evaluate`, so it is skipped too
-    # whenever this gate is false.
+    # and only when the PR is opted in via the `blast-radius` label (see `on:`
+    # above). The label is matched per-event so adding an UNRELATED label to an
+    # already opted-in PR does not re-claim a self-hosted eval runner (the whole
+    # point of the opt-in): the workflow fires on every `labeled` event, so the
+    # last clause requires that on a `labeled` event the label just added
+    # (`github.event.label.name`) is `blast-radius`, while `synchronize`/
+    # `reopened` (which carry no `label` payload) gate on the label still being
+    # present. The `comment` job has no `if:` of its own but `needs: evaluate`,
+    # so it is skipped too whenever this gate is false.
     if: >-
       github.event_name == 'pull_request' &&
       github.event.pull_request.head.repo.full_name == github.repository &&
       github.event.pull_request.user.login != 'dependabot[bot]' &&
-      contains(github.event.pull_request.labels.*.name, 'blast-radius')
+      contains(github.event.pull_request.labels.*.name, 'blast-radius') &&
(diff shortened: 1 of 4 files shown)


andrewgazelka:

PR #1440, opened 2026-06-20
ci(blast-radius): fix the empty PR comment end-to-end (opt-in trigger + resilient comment job + ?/= charset)

andrewgazelka:
## Why

A teammate reported the blast-radius PR comment "sometimes comes up empty." The rendered body is **never blank** (even a 0-rebuild report renders the `### Blast radius` header), so "empty" means the sticky comment is **absent / not updated**. Three independent root causes, validated against live `main`:

1. **Trigger disabled on PRs** (`#1384`) — `pull_request` was commented out for shared-runner cost, so the comment never ran. Re-enabled **opt-in** via the `blast-radius` label (`labeled`/`synchronize`/`reopened`), respecting #1384: unlabeled PRs claim no runner.
2. **Validator charset rejects `?`/`=`** (closes #1421) — `name_ok`/`safename` used `^[A-Za-z0-9_./ +():-]+$`, but cause names are nix derivation names and fixed-output patch/fetch drvs legally carry `?`/`=` (`<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). One such name fail-closed the **whole** report → no comment. Widened to the complete legal nix-name set `^[A-Za-z0-9_./ +()?=:-]+$` (still rejects every mention/HTML/markdown glyph). Rust mirror doc in `nix.rs` updated lockstep.
3. **Comment job skipped on eval bail** (closes #1415) — `comment` had a default `success()` gate, so a base-side/transient eval failure skipped it entirely. Now `if: !cancelled() && needs.evaluate.result != 'skipped'` with a "Compose fallback comment" step that posts an explicit "could not compute … transient base/runner issue" note under the same `<!-- blast-radius -->` marker (overwritten in place by the next good run). `Post sticky comment` keeps its fail-closed `success()` gate.

## Test

`tools/blast-radius-test.sh` (the `blast-radius-test` flake check, `lib/per-system.nix`) — extracts the real validate/render/fallback run-scripts from this workflow and asserts them against fixtures. `special-name.json` covers the `?`/`=` regression; a fallback fixture covers #1415. All 16 assertions pass locally.

## History

This is the consolidated end-to-end fix. A swarm of parallel agent runs opened ~10 overlapping PRs (#1413/#1416/#1422/#1424/#1425/#1426/#1427/#1428/#1430/#1432) that **cross-closed each other in a circular supersession** (#1428 → "see #1430", #1430 → "see #1428", both closed), so the fix never landed. Reopened to break the deadlock and drive a single PR to merge.

> [!NOTE]
> ### Fix empty PR comments in blast-radius CI with opt-in label trigger and fallback comment
> - The `blast-radius` workflow now triggers on `pull_request` events only when the PR carries the `blast-radius` label, replacing the previously disabled trigger.
> - The comment job posts a fallback 'Could not compute' sticky comment when evaluation fails, instead of silently producing no comment.
> - Broadens the validator and renderer regex to accept `?` and `=` characters in cause names (e.g. `?full_index=1`), fixing schema validation failures for these names.
> - Adds a new test fixture and regression tests in [blast-radius-test.sh](https://github.com/indexable-inc/index/pull/1440/files#diff-008b7e349202f69865582328028ef88fb2c6c22cdfd1f42fbbf5533600c2e27b) covering the special-character and fallback paths.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9b72044.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -48,15 +48,20 @@ jobs:
   evaluate:
     # Same-repository, non-Dependabot PRs only (mirroring ai-review-gate.yml),
     # and only when the PR carries the `blast-radius` label (the opt-in gate;
-    # see `on:` above). `labeled` events already carry the new label in
-    # `pull_request.labels`, so `contains` covers both the initial label-add and
-    # every later `synchronize`/`reopened` while the label remains. The `comment`
-    # job has `needs: evaluate` so it is skipped too whenever this gate is false.
+    # see `on:` above). `contains` covers the steady state for every event while
+    # the label remains. The final clause keeps a `labeled` event from running
+    # the expensive eval when the label being added is something *other* than
+    # `blast-radius`: without it, labelling an already opted-in PR with any
+    # unrelated label would re-claim the self-hosted runner and undercut the
+    # opt-in. `synchronize`/`reopened` (action != 'labeled') still refresh the
+    # comment via the `contains` check alone. The `comment` job has
+    # `needs: evaluate` so it is skipped too whenever this gate is false.
     if: >-
       github.event_name == 'pull_request' &&
       github.event.pull_request.head.repo.full_name == github.repository &&
       github.event.pull_request.user.login != 'dependabot[bot]' &&
-      contains(github.event.pull_request.labels.*.name, 'blast-radius')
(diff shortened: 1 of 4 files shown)


andrewgazelka:

andrewgazelka:

PR #1442, opened 2026-06-20
fix(blast-radius): restore the PR comment end-to-end (charset #1421 + eval-failure fallback #1415 + opt-in re-enable)

andrewgazelka:
## Problem

A teammate reported the blast-radius PR comment **"sometimes comes up empty"** on index PRs. Reproduced and root-caused from live state before touching anything. A *posted* comment is never literally blank (the renderer always emits a header), so **"empty" means the comment is absent.** Three independent causes can each drop it; this one atomic change fixes all three.

## Root causes (all reproduced)

1. **Validator charset rejects legal nix names (#1421).** The trusted `comment` job validates every cause/check name against `^[A-Za-z0-9_./ +():-]+$` and **fail-closes the whole report** (`::error::malformed`, `exit 1` → no comment) if any one name is outside it. But cause names are **nix derivation names**, and fixed-output patch/fetch drvs legally carry `?` and `=` (`<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). One such name in the rebuild frontier drops the whole comment, so it reproduces only "sometimes" (data-dependent) and only in CI (the Rust producer has no such guard). **Verified red→green:** feeding the current main validator the #1421 MRE → `exit 1`; widened → `exit 0`; an injection name (`@everyone \`id\``) still → `exit 1`.
2. **Eval bail silently skips the comment (#1415).** `comment` was `needs: evaluate` with the default `if: success()`, so a transient `evaluate` failure (often a base-side `main` eval, unrelated to the PR) skipped the comment entirely.
3. **Trigger disabled (#1384).** The `pull_request` trigger was commented out to stop every PR claiming a shared `ix-ci` runner slot — correct for runner load, but it left **no comment at all**.

## Fix

- **Charset:** widen `name_ok` (validator) and `safename` (renderer), kept lockstep, to `^[A-Za-z0-9_./ +()?=:-]+$` — the legal nix-name glyphs that reach a label. Still rejects every markdown/mention/HTML glyph. Rust `normalize_attr` doc updated to match.
- **Resilience:** `comment` now runs on eval success *or* failure (`if: !cancelled() && needs.evaluate.result != 'skipped'`) and posts an explicit "could not compute…" note under the same `<!-- blast-radius -->` marker, so a later good run overwrites it in place. Every gated step restates `success()` so the validate→render→post chain stays fail-closed.
- **Opt-in re-enable:** trigger on `pull_request: [labeled, synchronize, reopened]` gated on a `blast-radius` label, so the report is reachable again without re-introducing the #1384 every-PR runner cost. Unlabeled PRs claim no runner.

## Tests

`tools/blast-radius-test.sh` (the `blast-radius-test` flake check, wired in `lib/per-system.nix`) extends the suite with the `?`/`=` charset regression (validate + render), the fallback-comment path, and the fail-closed gate wiring. **All 20 assertions pass locally** (`nix shell nixpkgs#yq-go nixpkgs#jq -c bash tools/blast-radius-test.sh`); `actionlint` clean on the workflow.

Closes #1421
Closes #1415

## Note on the duplicates

A swarm of parallel agent runs opened ~12 overlapping PRs for this (#1413, #1416, #1422, #1424–#1428, #1430–#1433) that closed each other as "duplicates" in a circle — none ever merged. This is the single consolidated PR; the rest are closed and their branches will be cleaned up.

> [!NOTE]
> ### Restore blast-radius PR comment posting with label gating and eval-failure fallback
> - The [blast-radius.yml](.github/workflows/blast-radius.yml) workflow now triggers on PR events only when a maintainer applies the `blast-radius` label, skipping evaluation for Dependabot and fork PRs.
> - When the evaluate job fails, the comment job now posts a minimal fallback 'could not compute' note instead of silently posting nothing.
> - The jq schema validator and renderer now accept `?` and `=` in cause names, allowing valid Nix derivation names to pass validation; a new fixture [special-name.json](https://github.com/indexable-inc/index/pull/1442/files#diff-fd11772d46c6fe1a44e23e9db9179c9569dfcc29e4cd10108f2ea8db91281c1c) covers this case.
> - [blast-radius-test.sh](https://github.com/indexable-inc/index/pull/1442/files#diff-008b7e349202f69865582328028ef88fb2c6c22cdfd1f42fbbf5533600c2e27b) adds tests for `?`/`=` name handling, fallback comment generation, and fail-closed gating assertions on workflow steps.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 55e8dbf.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,17 +1,22 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in per PR, not on every push. Auto-running on every PR meant the per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claimed a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, queuing an
+  # eval-only runner each push and slowing the queue (#1384). But fully disabling
+  # it left the report unreachable, so the sticky comment "came up empty" on every
+  # PR. The middle ground: it runs only when a maintainer adds the `blast-radius`
+  # label. Once labeled, each later push (`synchronize`) and reopen refreshes the
+  # comment. The job `if:` below gates EVERY event on the label, so adding any
+  # other label (or pushing to an unlabeled PR) skips `evaluate` and claims no
+  # runner. `workflow_dispatch` is kept for manual reruns (no-ops without PR
+  # context).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


andrewgazelka:

andrewgazelka:

andrewgazelka:
Superseded by #1431, which carries the same end-to-end fix (charset #1421 + eval-failure fallback #1415 + opt-in re-enable) plus concurrency-cancel and unrelated-label guards, and passes `tools/blast-radius-test.sh` (20/20). Landing #1431 to avoid the duplicate-PR thrash. (sent by an AI agent via Claude Code)

PR #1444, opened 2026-06-20
ci(blast-radius): restore the PR comment end to end (opt-in label, eval-failure fallback, nix-name charset)

andrewgazelka:
## What

Restore the **blast-radius PR comment** end to end. A teammate reported it "sometimes comes up empty." Reproduced and root-caused before touching anything; the rendered body is never blank, so **"empty" means the comment is absent.** Three independent causes each drop it:

1. **Trigger disabled (#1384).** `pull_request` was commented out for runner cost, so PRs get **no comment at all** (dominant cause today).
2. **Eval bail silently skips the comment.** `comment` was `needs: evaluate` + default `if: success()`; a (frequently base-side, transient) eval failure dropped the comment entirely.
3. **Validator charset rejects legal nix names ("sometimes").** Both the schema validator and renderer matched `^[A-Za-z0-9_./ +():-]+$`, but cause names are **nix derivation names** whose legal charset includes `?` and `=` (e.g. `<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). One such name in the rebuild frontier `::error::malformed`-ed the whole report → exit 1 → no comment. Only fires when the frontier hits such a drv, hence "sometimes", and only in CI (the Rust producer has no such guard).

## Fix

- **Re-enable on PRs, opt-in via a `blast-radius` label** (respects #1384's runner-cost reason). `types: [labeled, synchronize, reopened]`; `evaluate` gated on the label, so unlabeled / fork / Dependabot PRs claim no runner.
- **Resilient `comment` job.** Runs on eval success **or** failure (`!cancelled() && needs.evaluate.result != 'skipped'`); download/validate/render stay gated on success; on failure a new step posts an explicit "could not compute" sticky note (same `<!-- blast-radius -->` marker, overwritten by the next good run). Post stays fail-closed (default `success()`), so a half-written body is never published.
- **Widen `name_ok` + `safename` in lockstep** to `^[A-Za-z0-9_./ +()?=:-]+$`. Still rejects every markdown / mention / HTML glyph (`@ < > [ ] \` & | ; * # {`, newline). `nix.rs` doc comment updated to match.

## Tests

`tools/blast-radius-test.sh` (the `blast-radius-test` flake check) extracts the workflow's jq/steps verbatim so it can't drift from CI. Adds a **special-name** fixture (cause name with `?`/`=`) and a **fallback-path** case. Verified locally: **17/17 pass on this branch**; main's validator emits `::error::malformed` (exit 1) on the same legal name (red→green).

### Security
Re-enabling `on: pull_request` runs the write-token `comment` job from PR YAML, bounded to same-repo PRs; forks get a read-only token and are excluded by the job `if:`. Hardening the post phase into a `workflow_run` context is a tracked follow-up.

Supersedes the closed churn #1413/#1416/#1422/#1424/#1425/#1426/#1427/#1428/#1430/#1431/#1432/#1433 (a prior agent loop) and collapses them into this one verified change.

> [!NOTE]
> ### Restore blast-radius PR comment with opt-in label, eval-failure fallback, and Nix name charset fix
> - The [blast-radius.yml](.github/workflows/blast-radius.yml) workflow now triggers on `labeled`, `synchronize`, and `reopened` PR events, gated by the `blast-radius` label.
> - Adds a fallback sticky comment step that posts an explanation with a link to the run URL when the `evaluate` job fails, instead of silently doing nothing.
> - Expands the safename regex in the validator and renderer to accept `?` and `=` characters, which are legal in Nix derivation names; adds a [test fixture](tools/blast-radius-fixtures/special-name.json) and assertions in [blast-radius-test.sh](tools/blast-radius-test.sh) to cover this.
> - The comment job now runs on both `evaluate` success and failure, but skips when the job is cancelled or skipped.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0bf6f1a.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,17 +1,21 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in on PRs via a `blast-radius` label, not every PR. The per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claims a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, so running
+  # it on every PR slowed the queue -- why #1384 disabled the trigger entirely.
+  # Gating on the label (see the `evaluate` job `if:`) keeps that cost off
+  # unlabeled PRs while restoring the comment for PRs a maintainer opts in.
+  # `labeled` fires the first run when the label is added; `synchronize` /
+  # `reopened` refresh the sticky comment on later pushes (each re-checks the
+  # label, so removing it stops future runs). `workflow_dispatch` keeps `on:`
+  # non-empty for manual runs (PR-context gated, so a bare dispatch no-ops).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Closing as a duplicate of #1428, which is the consolidated, verified fix for the same teammate report (empty blast-radius PR comment). #1428 is a strict superset: opt-in `blast-radius` label trigger + labeled-event guard, eval-failure fallback comment, widened nix-name charset (`?`/`=`), the Rust `normalize_attr` doc mirror, and the fuller `tools/blast-radius-test.sh` suite (all assertions pass locally, actionlint clean). Consolidating to avoid duplicate PRs.

andrewgazelka:
Superseded by #1445 (the consolidated canonical fix). Its more-robust fallback covers render/validate failures too, and I grafted this PR's run link + `nix.rs` charset doc onto it.

Issue #1429, opened 2026-06-20
blast-radius PR comment still "comes up empty" when eval succeeds but the comment pipeline flakes


andrewgazelka:
### Symptom

After indexable-inc/index#1411/indexable-inc/index#1415 (opt-in label gate + fallback on eval failure), the blast-radius PR comment can still **come up empty** on labeled PRs.

### Root cause (comment job step state machine)

The `comment` job's fallback fires only on `needs.evaluate.result != 'success'`. But when `evaluate` **succeeds** and the comment-side pipeline flakes, no comment is produced:

1. `Download report` flakes (transient `actions/download-artifact`) — `continue-on-error` protects only that step.
2. `Validate report schema` then errors on the missing/bad `report.json` (no `continue-on-error`) → job has a failed step.
3. `Render comment` is gated only on `needs.evaluate.result == 'success'` (an explicit `if:` replaces the default `success()`), so it **also** runs and errors.
4. `Compose fallback` is gated `needs.evaluate.result != 'success'` → **skipped** (eval succeeded).
5. `Post sticky comment` uses default `success()` → **skipped** (validate/render failed).

Net: evaluate green, but no comment and no fallback. Same silent-empty class as indexable-inc/index#1415, different trigger (eval OK, comment pipeline failed).

A simulation of the step state machine confirms `eval ok + download flake` and `eval ok + malformed report` both yield no post.

### Fix

Guarantee a comment whenever `evaluate` ran (not skipped/cancelled):

* `continue-on-error: true` on validate + render so transient/fail-closed failures don't poison `success()`.
* Gate render on `steps.validate.outcome == 'success'` (don't render a rejected report; preserves fail-closed for hostile reports).
* Atomic render write (no partial `comment.md`).
* Fallback fires on `needs.evaluate.result != 'success' || steps.render.outcome != 'success'`.
* Generalize fallback copy to cover eval-failure **and** transient/render failure.

### Test

The existing `tools/blast-radius-test.sh` exercises the extracted shell scripts but not the `if:` orchestration where this bug lives. Add orchestration invariants so the guarantee can't silently regress.

andrewgazelka:
Confirmed still reproduces against the current canonical PR #1431 (tip `4b589e85`): its comment job uses `&& success()` on Validate/Render and fires the fallback only on `needs.evaluate.result != 'success'`, so an eval-OK run whose `Download`/`Validate` flakes leaves no comment and no fallback (full trace in the #1431 review comment).

[...]

PR #1446, opened 2026-06-20
fix(blast-radius): stop the PR comment coming up empty (charset #1421 + resilient job #1415 + opt-in re-enable #1384)

andrewgazelka:
## Why
A teammate reported the blast-radius PR comment **"sometimes comes up empty"** on index PRs. A *posted* comment is never literally blank (the renderer always emits a header), so "empty" = **absent**.

This is the single consolidated, fully-validated fix. Many duplicate branches/PRs were opened for this during investigation; they are superseded by this one and will be closed.

## Root causes (all reproduced)
1. **Validator charset rejects legal nix names (#1421).** The trusted `comment` job validates every cause/check name against `name_ok` and **fail-closes the whole report** (exit 1 → no comment) if any name is outside it. Cause names are nix derivation names; fixed-output patch/fetch drvs legally carry `?` and `=` (`<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`), which the old set `^[A-Za-z0-9_./ +():-]+$` omitted. Confirmed: feeding the old validator such a report exits 1.
2. **Comment job skipped when eval bails (#1415).** `comment` had `needs: evaluate` + default `if: success()`, so a base/head eval failure (often base-side, transient, unrelated to the PR) silently skipped it — no sticky comment posted or refreshed.
3. **Disabled on PRs (#1384).** The `pull_request` trigger was commented out for runner cost; the comment was unreachable on every PR.

## Fix
- **Charset:** widen `name_ok` and the renderer `safename` (kept lockstep) to `^[A-Za-z0-9_./ +()?=:-]+$` — the complete legal nix store-path-name set plus attr-path glyphs. Still rejects every markdown/mention/HTML injection glyph (`@ < > [ ] \` & | ; * # { "`). Rust `normalize_attr` doc updated to match.
- **Resilience:** `comment` runs on eval success **or** failure (`if: !cancelled() && needs.evaluate.result != 'skipped'`); on failure it posts an explicit "could not compute … transient base/runner issue" note under the same `<!-- blast-radius -->` marker, overwritten in place by the next good run. Every gated step re-states `&& success()` so the validate→render fail-closed chain still holds.
- **Re-enable, opt-in:** trigger on `pull_request: [labeled, synchronize, reopened]` gated on the `blast-radius` label; label-add events are further narrowed to the `blast-radius` label itself so adding any *other* label to an opted-in PR does not re-claim a self-hosted eval runner. Respects #1384.

## Tests
`tools/blast-radius-test.sh` (the `blast-radius-test` CI check) — 21 assertions, all pass locally (charset regression via new `special-name.json`, fallback-comment path, fail-closed gate wiring, opt-in label-add narrowing). `actionlint` clean.

Closes #1415
Closes #1421

> [!NOTE]
> ### Fix blast-radius PR comment posting empty by adding fallback comment and special-char support
> - The workflow now triggers on `labeled`, `synchronize`, and `reopened` PR events, but only runs the expensive evaluate job when the PR has the `blast-radius` label; other label additions are ignored.
> - The comment job now posts an explicit fallback comment with a run link when evaluation fails, instead of producing no output.
> - Schema validation and rendering now accept cause names containing `?` and `=` characters, preventing valid Nix derivation names from being rejected.
> - Tests in [blast-radius-test.sh](https://github.com/indexable-inc/index/pull/1446/files#diff-008b7e349202f69865582328028ef88fb2c6c22cdfd1f42fbbf5533600c2e27b) now cover special-name acceptance, fallback comment behavior, and gating correctness.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6d2700d.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,17 +1,22 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in per PR, not on every push. Auto-running on every PR meant the per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claimed a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, queuing an
+  # eval-only runner each push and slowing the queue (#1384). But fully disabling
+  # it left the report unreachable, so the sticky comment "came up empty" on every
+  # PR. The middle ground: it runs only when a maintainer adds the `blast-radius`
+  # label. Once labeled, each later push (`synchronize`) and reopen refreshes the
+  # comment. The job `if:` below gates EVERY event on the label, so adding any
+  # other label (or pushing to an unlabeled PR) skips `evaluate` and claims no
+  # runner. `workflow_dispatch` is kept for manual reruns (no-ops without PR
+  # context).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Superseded by #1449, which consolidates the complete end-to-end fix (charset #1421 + opt-in label trigger #1411 + resilient/fail-closed comment job covering both eval failure #1415 and the comment-pipeline-flake gap #1429, with orchestration tests). Closing this duplicate.

PR #1448, opened 2026-06-20
ci(blast-radius): restore the PR comment end-to-end (charset + resilient + opt-in)

andrewgazelka:
## Why

A teammate reported the blast-radius PR comment **"sometimes comes up empty"** on index PRs. Investigating from live state surfaced **three independent root causes**, each filed as its own issue. This is the single consolidated fix (supersedes the duplicate attempts #1428 / #1430 / #1431).

## Root causes (all verified)

1. **Validator rejects legal nix names (#1421) — the "sometimes" symptom.** A cause name is a derivation name, and fixed-output patch/fetch drvs legally carry `?` and `=` (e.g. `<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`; both exist in the live store). The trusted comment job validates every name against a charset that omitted those two glyphs and **fail-closes the whole report**, so the comment job exits 1 and posts nothing whenever a rebuild frontier touches such a drv. **MRE:** feeding the current `main` validator a report with a `…patch?full_index=1` cause name returns `false` / exit 1.
2. **Comment job silently skipped when eval bails (#1415).** `comment` was `needs: evaluate` with the default `if: success()`, so a transient base-side `main` eval failure (unrelated to the PR) silently skipped the comment entirely.
3. **Trigger fully disabled (#1411, after #1384).** #1384 commented out `on: pull_request` to stop per-PR evals claiming a shared `ix-ci` runner, which left the comment unreachable on every PR.

## Fix

- **Charset:** widen `name_ok` / `safename` (kept lockstep) to the legal nix-name set `^[A-Za-z0-9_./ +()?=:-]+$`. Still rejects every markdown/mention/HTML glyph (`@ < > [ ] \` & | ; * # {`). New `special-name.json` fixture + regression test.
- **Resilient comment job:** runs on eval failure too and posts an explicit "could not compute" note under the same `<!-- blast-radius -->` marker, so the next good run overwrites it. The produce→post chain stays **fail-closed**: every gated step re-states `success()`, so a malformed report or half-written `comment.md` is never published (asserted in the test).
- **Opt-in re-enable:** trigger on `pull_request: [labeled, synchronize, reopened]` gated on the `blast-radius` label, so it claims a runner only when a maintainer opts in (preserving #1384's load-shedding intent). `labeled` events are gated on the `blast-radius` label itself, so adding an unrelated label to an opted-in PR does not restart the expensive eval.

## Security note (deliberate, see follow-up)

Re-enabling `on: pull_request` means the write-token `comment` job runs shell defined in the PR's own copy of this workflow file. This is the **pre-#1384 posture** and is an accepted residual risk here: the `evaluate` job already runs arbitrary PR-authored Nix (`nix run .#blast-radius`) on the self-hosted `ix-ci` runner, so anyone who can trigger this feature (same-repo PR **and** a maintainer-applied label) already has far more than the least-privilege `pull-requests: write` comment token. Hardening the comment job onto a base-trusted `workflow_run` trigger is tracked as a follow-up (that pattern cannot be validated until it is on the default branch, so it is intentionally out of scope for this bug fix).

## Test

`tools/blast-radius-test.sh` (wired as the `blast-radius-test` flake check) — **20/20 pass locally**, including the new `special-name` (charset), `fallback` (eval-failed), and fail-closed gate-wiring cases. `actionlint` clean. The `pull_request` trigger runs the PR's own workflow file, so I will validate the live comment by labelling this PR.

Closes #1411
Closes #1415
Closes #1421

> [!NOTE]
> ### Restore blast-radius PR comment with opt-in label gating, fallback comment, and extended charset support
> - The [blast-radius workflow](.github/workflows/blast-radius.yml) is now opt-in: it only runs on PRs labeled `blast-radius`, plus subsequent `synchronize`/`reopened` events, replacing unconditional push-based triggers.
> - When evaluation fails, a new fallback step posts a minimal comment to the PR explaining the blast radius could not be computed and linking to the run logs, instead of leaving the PR with no comment.
> - The report validator and renderer now accept cause names containing `?` and `=` characters (e.g. Nix derivation names with query parameters), fixing silent rejection of valid reports.
> - A new fixture [`special-name.json`](https://github.com/indexable-inc/index/pull/1448/files#diff-fd11772d46c6fe1a44e23e9db9179c9569dfcc29e4cd10108f2ea8db91281c1c) and tests in [`blast-radius-test.sh`](https://github.com/indexable-inc/index/pull/1448/files#diff-008b7e349202f69865582328028ef88fb2c6c22cdfd1f42fbbf5533600c2e27b) cover the extended charset, fallback comment generation, and fail-closed gating on each step.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 50169a9.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,17 +1,22 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in per PR, not on every push. Auto-running on every PR meant the per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claimed a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, queuing an
+  # eval-only runner each push and slowing the queue (#1384). But fully disabling
+  # it left the report unreachable, so the sticky comment "came up empty" on every
+  # PR. The middle ground: it runs only when a maintainer adds the `blast-radius`
+  # label. Once labeled, each later push (`synchronize`) and reopen refreshes the
+  # comment. The job `if:` below gates EVERY event on the label, so adding any
+  # other label (or pushing to an unlabeled PR) skips `evaluate` and claims no
+  # runner. `workflow_dispatch` is kept for manual reruns (no-ops without PR
+  # context).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Superseded by #1449, which consolidates the complete end-to-end fix (charset #1421 + opt-in label trigger #1411 + resilient/fail-closed comment job covering both eval failure #1415 and the comment-pipeline-flake gap #1429, with orchestration tests). Closing this duplicate.

PR #1449, opened 2026-06-20
fix(blast-radius): stop the PR comment coming up empty (charset + opt-in + resilient, end-to-end)

andrewgazelka:
## Why

A teammate reported the blast-radius PR comment **"sometimes comes up empty"** on index PRs. Investigating from live run history (the posted comment body is *never* blank — even a 0-rebuild report renders the `### Blast radius` header), **"empty" means the sticky comment is absent.** Three independent causes, fixed together so the comment is reliable end to end. This **supersedes the duplicate fix PRs #1428 / #1430 / #1431** (each fixed a subset); those are being closed in favor of this one.

## Root causes (all validated)

1. **Validator charset rejected legal nix names (#1421) — the actual "sometimes".** The trusted `comment` job validates every cause/check name against `name_ok` and *fail-closes the whole report* if any name is outside it. Cause names are nix derivation names, and fixed-output patch/fetch drvs legally carry `?` and `=` (`<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). A PR whose rebuild frontier touched such a drv made the validate step exit 1 → comment job failed → no comment. **MRE:** the old regex `^[A-Za-z0-9_./ +():-]+$` rejects `...patch?full_index=1`; the widened `^[A-Za-z0-9_./ +()?=:-]+$` accepts it (and `abi=4.1`) while still rejecting every markdown/mention/HTML glyph (`@ < > [ ] ` `` ` `` ` & | ; * # {`).
2. **Disabled on PRs (#1411 / #1384).** #1384 commented out the `pull_request` trigger (per-PR eval claimed a shared `ix-ci` runner), so the comment never appears.
3. **Comment job skipped / still empty on a flaked pipeline (#1415 / #1429).** The job was `needs: evaluate` with the default `if: success()`, so an eval bail (often base-side, unrelated to the PR) skipped it. #1429 is subtler: even when eval *succeeds*, a download flake or rejected report left no comment **and** no fallback.

## Fix

- **Charset:** widen `name_ok` + the renderer's `safename` in **lockstep** to the legal nix-name set; `nix.rs` `normalize_attr` doc updated to match. Fixture `special-name.json` + regression test.
- **Opt-in trigger:** re-enable `pull_request` (`labeled` / `synchronize` / `reopened`) gated on the `blast-radius` label. The `evaluate` `if` re-checks the label every event and ignores a `labeled` event for any *other* label, so unlabeled pushes claim no runner — **respects #1384**.
[...]

PR #1453, opened 2026-06-20
ci(blast-radius): stop the PR comment coming up empty (opt-in trigger + nix-name charset + resilient fallback)

andrewgazelka:
## What

The blast-radius sticky PR comment "sometimes comes up empty". "Empty" is actually **absent** (the renderer always emits a header). Three independent, reproduced causes, all still live on `main`:

1. **Trigger disabled (#1384).** `on: pull_request` is commented out, so the workflow never runs on PRs -> no comment on any PR. Re-enabled **opt-in via the `blast-radius` label** (`types: [labeled, synchronize, reopened]`, evaluate `if:` gates on the label), which keeps the per-PR full-catalog eval off the shared `ix-ci` runner pool unless a maintainer opts in -- respecting why #1384 disabled it.
2. **Comment skipped on eval failure (#1415/#1429).** The `comment` job had a default `success()` gate, so a transient (often base-side) eval bail skipped it silently. Now the job runs on `!cancelled() && evaluate != 'skipped'`; download/validate/render are `continue-on-error` and gated on prior outcomes (fail-closed: render only on a validated report), and a **fallback "could not compute" note** posts under the same marker when no usable report exists.
3. **Validator charset rejects legal nix names (#1421).** `name_ok`/`safename` matched `^[A-Za-z0-9_./ +():-]+$`, but cause names are nix derivation names that legally carry `?`/`=` (`<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). One such name fail-closed the whole report -> no comment, only "sometimes". Widened both (lockstep) to `^[A-Za-z0-9_./ +()?=:-]+$`, still rejecting every injection glyph. Rust mirror doc in `nix.rs::normalize_attr` updated.

## Test

`tools/blast-radius-test.sh` (the `blast-radius-test` flake check) extracts the workflow's validate/render/fallback scripts verbatim so they can't drift, plus `special-name.json` (the `?`/`=` regression), overflow cap, byte backstop, fallback, and structural orchestration invariants (continue-on-error + fail-closed gating). All 22 assertions pass locally.

## Context

Supersedes the large cluster of duplicate empty-comment PRs (#1428/#1431/#1442/#1444-1449 etc.) created by parallel agents racing the same fix. This is one clean, validated consolidation off current `main`.

Gate: only `flake-check` is required. ai-review checks are advisory (fail org-wide).

> [!NOTE]
> ### Fix blast-radius CI comment posting via opt-in label trigger, extended charset support, and fallback comment
> - The workflow now triggers on `pull_request` events only when the PR carries a `blast-radius` label, replacing the previous always-on behavior; `workflow_dispatch` is retained.
> - The comment job posts a fallback "Could not compute the blast radius" comment (with a run log link) when the report is missing, malformed, or rendering fails, instead of posting nothing.
> - Cause names containing `?` and `=` (valid in nix derivation names) now pass schema validation and render correctly; a new fixture [`special-name.json`](https://github.com/indexable-inc/index/pull/1453/files#diff-fd11772d46c6fe1a44e23e9db9179c9569dfcc29e4cd10108f2ea8db91281c1c) covers this case.
> - [`tools/blast-radius-test.sh`](https://github.com/indexable-inc/index/pull/1453/files#diff-008b7e349202f69865582328028ef88fb2c6c22cdfd1f42fbbf5533600c2e27b) adds tests for the new charset, fallback behavior, and step orchestration invariants (validate gates render, fallback triggers on render failure).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1e280ab.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,17 +1,21 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in per PR, not on every push. Auto-running on every PR meant the per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claimed a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, queuing an
+  # eval-only runner each push and slowing the queue (#1384). But fully disabling
+  # it left the report unreachable, so the sticky comment "came up empty" on every
+  # PR. Middle ground: it runs only when a maintainer adds the `blast-radius`
+  # label. Once labeled, each later push (`synchronize`) and reopen refreshes the
+  # comment. The `evaluate` job `if:` below re-checks the label on every event, so
+  # pushing to an unlabeled PR (or adding a different label) claims no runner.
+  # `workflow_dispatch` is kept for manual reruns (no-ops without PR context).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
+    types: [labeled, synchronize, reopened]
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Superseded by #1449 (byte-identical/earlier variant of the same fix). Consolidating the duplicate blast-radius empty-comment PRs onto #1449, which has squash auto-merge enabled and passes tools/blast-radius-test.sh locally. Closing to cut the thrash.

andrewgazelka:
Closing in favor of #1464, the single consolidated fix for the empty blast-radius comment (charset `?`/`=` #1421, eval-failure fallback #1415, opt-in re-enable #1411). This PR was one of several parallel rollouts of the same task; only #1464 should land.

Issue #1423, opened 2026-06-20
blast-radius: harden the write-token comment job into a base-branch workflow_run context


andrewgazelka:
## Context

#1422 re-enables the blast-radius PR comment opt-in via the `blast-radius` label. Doing so runs the **write-token `comment` job from PR-controlled workflow YAML** (`on: pull_request` uses the workflow file from the PR merge commit). A same-repo PR could rewrite the `comment` job — drop the schema validation, change the posted body, etc. — and it would run with `pull-requests: write`.

## Why it's bounded today (not a fork-pwn)

- The job `if:` restricts to `head.repo.full_name == github.repository` (same-repo only).
- Fork PRs get a **read-only** `GITHUB_TOKEN` on `pull_request` regardless of `permissions:`.
- So the write-token path is reachable only by **collaborators who already have write access** to the repo — a weak threat model.

## Proposed hardening

Move the comment/post phase out of PR-controlled YAML into a base-branch trusted context:
- `evaluate` stays on `pull_request` (read-only token, runs PR `.#blast-radius` code, uploads the data-only `report.json` artifact).
- A separate `workflow_run`-triggered workflow (definition always taken from the default branch) downloads the artifact, validates the schema, renders, and posts with the write token.

Caveat for whoever picks this up: `workflow_run` triggers fire **only for the workflow version on the default branch**, so the posting workflow can't be exercised from a PR branch — it only activates after merge to `main`. Plan to validate by landing behind the existing `blast-radius` opt-in label and watching the first labeled PR.

sent by an AI agent (Claude) via Claude Code

andrewgazelka:
Closing: auto-filed by a runaway system-prompt eval rollout (live `claude -p` agents that ran tasks end-to-end), not a real request. Root cause fixed in #1471. (cleanup by an AI agent via Claude Code)

PR #1467, opened 2026-06-20
fix(blast-radius): stop the PR comment coming up empty (charset + opt-in + resilient)

### LINKED ISSUE #1429: blast-radius PR comment still "comes up empty" when eval succeeds but the comment pipeline flakes

andrewgazelka:
## Why

A teammate reported the blast-radius PR comment **"sometimes comes up empty."** The rendered body is never blank — **"empty" means the sticky comment is absent.** Validating from live run history surfaced multiple causes; this is the complete, tested end-to-end fix.

> Context: a burst of parallel agent sessions opened/closed ~10 overlapping fix PRs (#1428/#1430/#1431/#1446/#1447/#1448/#1449) for this same bug and then closed all of them, leaving `main` unfixed. This PR is the single consolidated, complete fix (it is also the only one that covers the #1429 residual). The duplicates are closed.

## Root causes (all validated)

1. **Validator charset rejected legal nix names (#1421) — the actual "sometimes".** The trusted `comment` job validates every cause/check name and *fail-closes the whole report* if any is outside `^[A-Za-z0-9_./ +():-]+$`. Cause names are nix derivation names; fixed-output patch/fetch drvs legally carry `?`/`=` (`<sha>.patch?full_index=1`, `webkitgtk-2.52.4+abi=4.1`). A PR whose rebuild frontier touched such a drv → validate exits 1 → no comment. **MRE:** the old regex rejects `...patch?full_index=1`; the widened `^[A-Za-z0-9_./ +()?=:-]+$` accepts it while still rejecting every markdown/mention/HTML glyph.
2. **Disabled on PRs (#1411 / #1384).** #1384 commented out the `pull_request` trigger, so the comment never appears.
3. **Comment job skipped / still empty on a flaked pipeline (#1415 / #1429).** It was `needs: evaluate` + default `if: success()` (skipped on eval bail); and even on a green eval, a download flake or rejected report left no comment *and* no fallback.

## Fix

- **Charset:** widen validator `name_ok` + renderer `safename` in **lockstep**; `nix.rs` doc updated; `special-name.json` fixture + regression test.
- **Opt-in trigger:** re-enable `pull_request` (`labeled`/`synchronize`/`reopened`) gated on the `blast-radius` label, ignoring `labeled` events for other labels, with a matching `concurrency.cancel-in-progress` guard — restores the comment without the per-PR runner cost #1384 removed.
- **Always-comment, fail-closed (#1415 + #1429):** comment job runs on eval failure too; `Download`/`Validate`/`Render` are `continue-on-error`, so eval bail, download flake, rejected report, or render hiccup all route to an explicit **"could not compute"** fallback (same marker, overwritten by the next good run). Fail-closed: `Render` is gated on `steps.validate.outcome == 'success'`, render write is atomic, `Post` keeps default `success()`.

## Tests

`tools/blast-radius-test.sh` (the `blast-radius-test` flake check) covers charset, fallback, overflow/backstop, and **yq orchestration invariants** for the continue-on-error / outcome-gating wiring. **All 21 assertions pass locally; `actionlint` clean.**

Follow-up: #1423 (move the write-token post into a base-branch `workflow_run` context) stays open as separate hardening.

Closes #1411
Closes #1415
Closes #1421
Closes #1429

> [!NOTE]
> ### Fix blast-radius PR comment being empty by adding opt-in label gating, charset fixes, and a fallback comment
> - The workflow now triggers on `labeled`, `synchronize`, and `reopened` PR events, but the evaluate job only runs for same-repo, non-Dependabot PRs that carry a `blast-radius` label — unrelated label events are ignored.
> - Artifact download, schema validation, and rendering steps now use `continue-on-error: true` so failures route to a fallback path instead of killing the job.
> - A new fallback step writes a minimal marker-prefixed comment when evaluation, download, validation, or rendering fails, ensuring the PR comment is never empty.
> - The jq regex in validation and rendering now accepts `?` and `=` in derivation names, fixing silent drops for paths like `?full_index=1`.
> - Behavioral Change: the comment job now runs even when `evaluate` fails (but still skips when `evaluate` was skipped), and concurrency cancellation no longer preempts in-progress runs triggered by unrelated labels.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bd02e2f.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -22,7 +22,15 @@ permissions:
 
 concurrency:
   group: blast-radius-${{ github.ref }}
-  cancel-in-progress: true
+  # Only cancel an in-flight run when the new event will itself run an eval: a
+  # push (`synchronize`)/`reopened`, a `workflow_dispatch`, or the `blast-radius`
+  # label being added. An unrelated `labeled` event skips `evaluate` (see the job
+  # `if:`), but concurrency cancellation is applied at the workflow level BEFORE
+  # any job `if:`, so an unconditional `cancel-in-progress` would let adding any
+  # label kill an opted-in PR's running eval and leave the sticky comment
+  # unrefreshed. Gating the cancel on the same condition keeps unrelated label
+  # events from interrupting a real run.
+  cancel-in-progress: ${{ github.event_name != 'pull_request' || github.event.action != 'labeled' || github.event.label.name == 'blast-radius' }}
 
 env:
   # The self-hosted runner's Nix daemon owns substituters; the job only sets
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,17 +1,21 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in per PR, not on every push. Auto-running on every PR meant the per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claimed a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, queuing an
+  # eval-only runner each push and slowing the queue (#1384). But fully disabling
+  # it left the report unreachable, so the sticky comment "came up empty" on every
+  # PR. Middle ground: it runs only when a maintainer adds the `blast-radius`
+  # label. Once labeled, each later push (`synchronize`) and reopen refreshes the
+  # comment. The `evaluate` job `if:` below re-checks the label on every event, so
+  # pushing to an unlabeled PR (or adding a different label) claims no runner.
+  # `workflow_dispatch` is kept for manual reruns (no-ops without PR context).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
+  pull_request:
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Closing: auto-generated by a runaway system-prompt eval rollout (the eval harness spawned live `claude -p` agents that opened PRs end-to-end), not an intended change. Root cause fixed in #1471. (cleanup by an AI agent via Claude Code)

PR #1468, opened 2026-06-20
ci(blast-radius): restore the PR comment end-to-end (opt-in label + nix-name charset + resilient fallback)

andrewgazelka:
## What

The blast-radius sticky PR comment was **coming up empty**. Root-caused to three independent failures, each reproduced and fixed:

1. **Trigger fully disabled** (#1384 commented out `pull_request`), so nothing posted at all. Re-enabled as **opt-in via the `blast-radius` label** — the expensive per-PR full-catalog eval only claims the shared `ix-ci` runner pool when a maintainer adds the label (keeps the cost win of #1384). Pushes/reopens refresh a labeled PR; the concurrency-cancel and job `if:` gates keep *unrelated* label events from triggering an eval.
2. **Charset bug** (#1421): the validator/renderer `name_ok`/`safename` regex rejected legal nix derivation names containing `?` and `=` — fixed-output patch/abi drvs like `<sha>.patch?full_index=1` and `webkitgtk-2.52.4+abi=4.1`. A single such cause fail-closed the *whole* report, so the comment job exited 1 and posted nothing. Widened the charset in lockstep (validator + renderer + `nix.rs` doc), still rejecting every markdown/mention/HTML glyph.
3. **Eval failure → no comment**: a transient base/head eval bail left the PR silent. Added a fail-closed *"could not compute"* fallback comment keyed by the same sticky marker, so the next good run overwrites it in place.

## Root-cause evidence (MRE)

Ran `main`'s exact validator jq against a report whose cause name is `<sha>.patch?full_index=1`:

```
FAIL rc=1 -> comment job would exit 1 and post nothing
```

After the charset widening it validates and renders with the name intact.

## Tests

`tools/blast-radius-test.sh` (wired as the `blast-radius-test` flake check) gains:
- **special-name fixture** — a `?`/`=` cause name now validates and renders intact (charset regression #1421).
- **fallback** — asserts the eval-failed comment carries the marker + run link.
- **gate-wiring** — asserts every gated comment-job step restates `success()` so the produce-then-post chain stays fail-closed (an explicit `if:` drops GitHub's implicit `success()`).

All 20 checks pass locally; `actionlint` clean.

## Note

This was produced as one of several parallel prompt-eval rollouts on the same task, so sibling PRs (e.g. #1431/#1449/#1463/#1464) carry overlapping fixes — this is the consolidated, tested version.

🤖 sent by an AI agent via Claude Code

> [!NOTE]
> ### Restore blast-radius PR comment with opt-in label gating and resilient fallback
> - The [blast-radius.yml](.github/workflows/blast-radius.yml) workflow now only runs evaluation on PRs carrying the `blast-radius` label, refreshing on subsequent pushes and reopens; unrelated label events no longer cancel in-flight runs.
> - When evaluation fails, the workflow still posts a sticky fallback comment with a marker and a link to the run, rather than silently doing nothing.
> - Extends the allowed character set for nix derivation names to include `?` and `=` in both the jq schema validator regex and the renderer, keeping them in lockstep.
> - Adds a [special-name.json](https://github.com/indexable-inc/index/pull/1468/files#diff-fd11772d46c6fe1a44e23e9db9179c9569dfcc29e4cd10108f2ea8db91281c1c) fixture and regression tests in [blast-radius-test.sh](https://github.com/indexable-inc/index/pull/1468/files#diff-008b7e349202f69865582328028ef88fb2c6c22cdfd1f42fbbf5533600c2e27b) covering `?`/`=` name acceptance, fallback comment contents, and fail-closed gating.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8254296.</sup>
> 
>

### DIFF
--- a/.github/workflows/blast-radius.yml
+++ b/.github/workflows/blast-radius.yml
@@ -1,24 +1,37 @@
 name: Blast radius
 
 on:
-  # Disabled on PRs: the per-PR full-catalog eval (~57s "Report blast radius"
-  # step) claims a shared `ix-ci` runner from the one dispatcher pool that
-  # serves the whole team, so every PR queued one more eval-only runner and
-  # slowed the queue. The report is informational (a sticky PR comment), never
-  # a merge gate, so dropping the auto-trigger costs no safety. Re-enable by
-  # uncommenting the `pull_request` block below. `workflow_dispatch` only keeps
-  # `on:` non-empty (the jobs are PR-context gated, so a manual run no-ops).
+  # Opt-in per PR, not on every push. Auto-running on every PR meant the per-PR
+  # full-catalog eval (~57s "Report blast radius" step) claimed a shared `ix-ci`
+  # runner from the one dispatcher pool that serves the whole team, queuing an
+  # eval-only runner each push and slowing the queue (#1384). But fully disabling
+  # it left the report unreachable, so the sticky comment "came up empty" on
+  # every PR. The middle ground: run only when a maintainer adds the
+  # `blast-radius` label. Once labeled, each later push (`synchronize`) and
+  # reopen refreshes the comment. The job `if:` gates EVERY event on the label,
+  # so adding any other label (or pushing to an unlabeled PR) skips `evaluate`
+  # and claims no runner. `workflow_dispatch` is kept for manual reruns (it
+  # no-ops without PR context).
   workflow_dispatch: {}
-  # pull_request:
-  #   branches:
-  #     - main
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Closing: auto-generated by a runaway system-prompt eval rollout (the eval harness spawned live `claude -p` agents that opened PRs end-to-end), not an intended change. Root cause fixed in #1471. (cleanup by an AI agent via Claude Code)

PR #1634, opened 2026-06-26
ci: publish index's public closure (incl. crate FODs) to cache.ix.dev

wyattgill9:
## Problem
`ix up` fails for fleets built from the base profile (masked as `internal server error`). The real cause: index's own build artifacts — notably the `nix-cargo-unit` / `ix-mcp` vendor **crate FODs** — are never published to any cache the fleet substitutes from. The build sandbox has no `static.crates.io` egress, so any crate not already warm in ncps 404s and the build dies. Today only `.#site` is pushed to the public cache; everything else (including build-time FODs) is absent.

## Change
Add a `main`-gated workflow on the self-hosted `ix-ci-run-*` runner that publishes index's **public build closure** to the `ix-public` bucket (`cache.ix.dev`):
- enumerates `.#packages.x86_64-linux` (recursively, derivations only),
- realizes them (warm store → near no-op),
- computes `nix-store --query --requisites --include-outputs` — **the `--include-outputs` is what captures the crate FODs**, not just runtime deps,
- publishes via the on-fabric `ix-public-cache-push` (nix copy → Garage S3, signed `ix-workspace`).

## Scope guard
Publishes **only `.#packages.x86_64-linux`** — index's public packages. It never touches `ciChecks`/tests or any private/fabric path. (Example-fleet `nixosConfigurations` toplevels are an easy add if you want their system closures warmed too — flagged for your call.)

## Review notes / prerequisites
- **Runner privilege:** `ix-public-cache-push` reads the secret-store env file and needs root. This assumes the `ix-ci-run-*` runner is granted `sudo` NOPASSWD for exactly that command. If you'd rather, it can run via a dedicated push credential instead of sudo — your call on the wiring.
- **Pairs with indexable-inc/ix#5802 / its PR**: that change fronts the `ix-public` bucket through ncps so fleets can *read* what this publishes. Neither fixes `ix up` alone; both are needed.
- Draft for review — does not run until merged to `main`, and self-contained (separate from the required `flake-check` status).

Part B of indexable-inc/ix#5802.

> [!NOTE]
> ### Publish index public closure including crate FODs to cache.ix.dev after CI checks
> Adds a new [warm-public-cache.yml](.github/workflows/warm-public-cache.yml) GitHub Actions workflow that runs after the 'Check' workflow succeeds on `main` (or on manual dispatch).
>
> - Evaluates and builds all package derivations for four target systems: `x86_64-linux`, `aarch64-linux`, `aarch64-darwin`, and `x86_64-darwin`
> - Computes full closures including build-time crate FODs via `nix-store --query --requisites --include-outputs`
> - Uploads resulting store paths to the public cache via `ix-public-cache-push`
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 053857e.</sup>
> 
>

### DIFF
--- a/.github/workflows/warm-public-cache.yml
+++ b/.github/workflows/warm-public-cache.yml
@@ -1,21 +1,11 @@
 name: Warm public cache
 
-# Publish index's public build closure to the ix-public flat cache
-# (cache.ix.dev) so fleet `ix up` builds and developer machines can substitute
-# index's own artifacts -- including build-time crate FODs (nix-cargo-unit /
-# ix-mcp vendor) that a build sandbox cannot fetch from crates.io. Companion to
-# the ix-repo change that fronts the ix-public bucket through ncps
-# (indexable-inc/ix#5802).
-#
-# Scope: PUBLIC outputs only -- `.#packages.<system>` for every system index
-# exposes. Never ciChecks/tests or any private path. The push goes through the
-# on-fabric `ix-public-cache-push` helper (nix copy -> Garage S3, signed with
-# the fleet `ix-workspace` key).
+# Publish index's public package closures (incl. build-time crate FODs) to the
+# ix-public bucket so fleet `ix up` and dev machines can substitute them.
 
 on:
   push:
-    branches:
-      - main
+    branches: [main]
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


wyattgill9:
Updated to address multi-system coverage (thanks for the catch): the publish now runs as a **matrix over all four `devSystems`** index exposes — `x86_64-linux`, `aarch64-linux`, `aarch64-darwin`, `x86_64-darwin` — not just x86_64-linux. `extra-system-features` (znver5) is per-system.

- x86_64-linux is what fleet `ix up` (Linux VMs) needs; aarch64-linux + the two Darwin entries warm the cache for ARM and **Mac developers** pulling index's dev tools.
- **Review note:** each non-x86_64-linux entry needs the ix-ci dispatcher to mint a runner of that arch/OS (today only x86_64-linux is wired for the check gate), *or* a single x86_64-linux runner with darwin/aarch64 remote builders to build all systems and publish from one host. `fail-fast: false` so a missing-runner arch doesn't drop the others.

wyattgill9:
Reworked for speed after checking what CI already does:

**Finding:** the check gate (`nix run .#check` → `nix-fast-build .#ciChecks.x86_64-linux --skip-cached`) already builds the full rust workspace + the `nix-cargo-unit`/`ix-mcp` vendor crate FODs on the warm self-hosted runner, incrementally. Building `.#packages` in a standalone workflow re-did that work and gave no guarantee the store was warm.

**Changes:**
- **Trigger off `Check` completing on `main`** (`workflow_run`) so the x86_64-linux store is already warm. The publish then realizes only the package *finals* the gate skips (e.g. `base` image) — **no workspace rebuild**.
- **Incremental by construction:** warm store ⇒ `nix build` is a no-op realize for the workspace; `nix copy` (in `ix-public-cache-push`) uploads only paths the bucket lacks, so repeat runs push just the diff.
- aarch64/darwin: nothing else builds them, so they build their (smaller) sets once and stay warm after.

**Two things for your call:**
1. `.#packages` includes `site` (already pushed by `pages.yml`) and `base`. Caching `base` helps `ix up`; `site` is redundant — want me to drop `site` from the publish set?
2. If you'd rather avoid even the second eval, the alternative is folding the push into the `check.yml` gate itself (it already evaluates everything) — more coupling to the required status, so I kept it separate. Say which you prefer.

PR #1739, opened 2026-07-02
panes: enable compositor gpu feature in the guest image (linux-dmabuf for venus WSI, #1686)

andrewgazelka:
Enables the panes-compositor `gpu` cargo feature in the guest image so venus Vulkan clients get a `linux-dmabuf` global — the Wayland-WSI-over-dma-buf half of the missing `VK_KHR_swapchain` (see the latest #1686 comment for the measured gap).

## What
- **`gpu` is now a default feature** of `panes-compositor`. The cargo unit graph (`lib/rust/cargo-unit.nix`) generates `cargo build --unit-graph` with no `--no-default-features` and has no per-package feature knob, so default features ARE the graph's feature selection; no build-system plumbing needed. The guest image is the sole consumer and needs the dmabuf path; GPU-less runs still degrade to shm-only at runtime (libEGL via libloading, no link-time GL dep).
- **Fix a latent compile error** in the never-CI-built gpu path (`Dmabuf::format` needs `smithay::backend::allocator::Buffer` in scope).
- **Put libglvnd on the compositor service's loader path**: smithay dlopens `libEGL.so.1` (the glvnd dispatcher), which nothing else provided; nixpkgs libglvnd defaults its vendor search to `/run/opengl-driver/share/glvnd/egl_vendor.d` and mesa's vendor JSON is store-absolute, so `LD_LIBRARY_PATH=libglvnd/lib` is the whole fix. Without it the gpu init panics-and-degrades to shm-only (observed on the first boot smoke).

## Validation (aarch64 builder VM + boot smoke on hydra)
- `nix build .#packages.aarch64-linux.panes-compositor` and `.#packages.aarch64-linux.panes-guest-image` via `ssh-ng://vm`: green.
- Booted the image with `vmkit boot-linux --disk /tmp/gpu-test.raw --gpu ...` (fresh copy, distinct vsock socket; live demo untouched). Console log:
  - `panes-compositor[966]: INFO panes_compositor::compositor: GPU readback ready; linux-dmabuf advertised`
  - venus intact: `deviceName = Virtio-GPU Venus (Apple M5 Max)`, `driverName = venus`, `/dev/dri/renderD128` present.
- `nix run .#lint`: 7/7 green.

## VK_KHR_synchronization2 investigation (report-only, the other missing extension)
Versions in the stack (nixpkgs pin `a799d3e3`, what vmkit's rebuilt libkrun-efi 1.19.3 links):
[...]

Issue #1686, opened 2026-07-01
Seamless Linux windows on macOS: per-container native windows via vmkit, Minecraft e2e


andrewgazelka:
Goal: seamless Linux windows on macOS, generic, Minecraft as the e2e proof.

One Linux VM (vmkit/libkrun, virtio-gpu Venus -> MoltenVK), N containers inside (declared in nix), one app per container. Each app's Wayland xdg_toplevel appears as a native-looking macOS window (NSWindow + CAMetalLayer), with input forwarded back. Target 120Hz on ProMotion.

Planned architecture:
- Guest: headless smithay-based compositor; apps (containers) connect as normal Wayland clients; per-toplevel frames streamed damage-driven over vsock (libkrun krun_add_vsock_port -> host unix socket); GL apps via Mesa zink on venus.
- Host: agent creating one NSWindow per remote toplevel, Metal blit, NSEvent -> wayland input backchannel.
- Containers: nix-declared spec; wayland socket + /dev/dri/renderD128 bind-mounted.
- Proof: Minecraft (Java, GL->zink->venus) playable in a native-looking window at 120Hz.

Milestones: M1 GPU guest + zink smoke; M2 guest compositor + vsock export; M3 host window agent; M4 nix container spec; M5 Minecraft e2e; M6 120Hz validation.

(sent by an AI agent via Claude Code)

andrewgazelka:
Big session update — PR #1714 is up with the full system. Validated tonight:

- **First real e2e windows**: foot + weston-flower render as native macOS windows over the real vsock leg (first ever test of that path)
- **120Hz genlock (M6)**: sustained **120.0 acks/s** measured on the real window path
- **Venus root-caused and FIXED (M1)**: `hv_vm_map` on Apple Silicon requires 16KiB-aligned size/addrs; a 4K-page guest kernel packs venus blobs 4K-aligned, so the first venus allocation fails (`0x1200` on MAP_BLOB/UNMAP_BLOB — NOT capset queries; 0x208/0x209 are map/unmap). Fix: `CONFIG_ARM64_16K_PAGES=y` guest kernel (muvm/Asahi configuration). vulkaninfo now shows `Virtio-GPU Venus (Apple M5 Max)`
- Containers fixed (missing `/nix/var/nix/db` for nspawn binds), clipboard global added (foot hard-requires `wl_data_device_manager`), texture race double-buffered

[...]

PR #1762, opened 2026-07-03
panes guest: bake the Vulkan startup switch into the MC launch (#1686)

andrewgazelka:
26.2 selects its startup renderer from the `--graphicsBackend vulkan` game argument (the official launcher passes it; joptsimple in Main) — the baked options.txt `preferredGraphicsBackend` seed alone boots GL, validated live during the venus bring-up. portablemc 5.0.3 has no game-arg passthrough, so the MC container's ExecStart becomes a checked wrapper (`writeBashApplication` via the guest overlay): first run prepares all downloads with `start --dry`, jq idempotently appends the pair to the version manifest's `arguments.game`, and the real launch passes `--fetch-exclude 26.2` so portablemc doesn't re-fetch the manifest over the injection.

Part of #1686 — the last piece of the GPU Minecraft path after #1739/#1741. Runtime validation on the real guest to follow on this PR.

🤖 Generated with Claude Code

> [!NOTE]
> ### Bake Vulkan backend flag into the Minecraft launch wrapper in panes guest
> - Adds `mcLaunch`, a bash wrapper in [apps.nix](https://github.com/indexable-inc/index/pull/1762/files#diff-a890cdfbf1cce913a32f1824d8badb40d6b06eb50aee87abc0bbdf1772968b64) that injects `--graphicsBackend vulkan` into the Minecraft version manifest's `.arguments.game` via `jq` before each launch, ensuring Vulkan is always used without manual configuration.
> - Self-heals missing or malformed version manifests by re-fetching with `portablemc start --dry` before launch; subsequent launches skip re-downloading via `--fetch-exclude`.
> - Centralizes portablemc arguments (JDK 25 path, LWJGL natives, Wayland flags, offline user) into a shared `mcStartArgs` binding to avoid duplication.
> - Exposes `writeBashApplication` through the overlay in [default.nix](https://github.com/indexable-inc/index/pull/1762/files#diff-5387531d78c5e6f1ef0e82ea770d5c1f36bbae27f41fe4da470808ee954cf0b7) to support the new wrapper.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5c7f398.</sup>
> 
>

### DIFF
--- a/packages/vm/panes/guest-image/apps.nix
+++ b/packages/vm/panes/guest-image/apps.nix
@@ -21,6 +21,8 @@
 #                           with baked config it can still rewrite at runtime
 { pkgs }:
 let
+  inherit (pkgs) lib;
+
   # portablemc's default wrapper bundles four full OpenJDKs (25/21/17/8,
   # ~1.9 GiB); its package.nix exposes `jvms` exactly to cut that closure.
   # MC 26.2 requires Java SE 25 minimum, so ship exactly jdk25 (the full JDK,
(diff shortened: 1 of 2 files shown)


andrewgazelka:
Review verdict: approve-with-fixes — all applied in 5c7f398a:
- **C1**: manifest guard is now parse-validity (`jq -e .`), not existence — a launch killed mid-write can no longer wedge the container in a never-re-downloading restart loop.
- **C2**: loud failure if `--dry` doesn't produce the manifest at the expected path.
- **M1**: `mcVersion` hoisted so the positional, manifest path, and `--fetch-exclude` move together on a bump.
- **M2**: coreutils pinned in runtimeInputs for `mv`.
- P1 (no-op jq rewrite each launch) declined as negligible per the reviewer's own note.

Reviewer also verified the flag grammar against portablemc 5.0.3's actual parse.rs (all flags valid, --fetch-exclude scope correct and safe unconditionally).

PR #1881, opened 2026-07-06
search-py: fix limit=0 panic across FFI, release GIL on BM25 work

### LINKED ISSUE #1880: file-search: limit=0 panics (surfaces as PanicException through search-py)
Post-merge adversarial review of #1879 found that a `limit=0` search panics in `file-search`, which pyo3 re-raises as a `PanicException` (a `BaseException`) across the FFI boundary.

## Root cause
`TopDocs::with_limit(0)` has `assert_ne!(limit, 0)` (tantivy 0.26.1). Any path that reaches it with `limit == 0` panics.

## Call chains
- `search-py` `bm25_rerank(query, [])` → `limit = texts.len() = 0` → `rerank` → `EphemeralSearch::search` → `packages/search/file-search/src/ephemeral.rs:96` `TopDocs::with_limit(0)`.
- `search-py` `bm25_rerank(..., limit=0)` / `bm25_search(..., limit=0)` → same panic via `ephemeral.rs` and `packages/search/file-search/src/search.rs:70`.
- The `file-search` CLI has the same latent panic via `--limit 0`.

## Fix
Fix at the source in `file-search`: early-return an empty result set when `limit == 0` (and for the empty-corpus rerank) instead of reaching `with_limit(0)`. One fix covers the CLI, the library, and the bindings.

Also a related performance issue: the three sync BM25 fns in search-py do all Tantivy CPU/disk work holding the GIL, freezing the caller's asyncio loop; they should wrap the blocking work in `py.allow_threads`.

Found by post-merge adversarial review of #1879.

andrewgazelka:
Follow-up to #1879, found by post-merge adversarial review.

- **C1/C2 panic across FFI**: `TopDocs::with_limit(0)` asserts `limit != 0` (tantivy 0.26.1). Any `limit=0` search panicked, and `bm25_rerank(query, [])` hit it via `limit = texts.len() = 0`; pyo3 re-raised it as a `PanicException`. Fixed at the source: `file-search`'s `search()` and `EphemeralSearch::search()` early-return empty on `limit == 0`, which also fixes the CLI `--limit 0` and the empty-corpus rerank in one place.
- **P1 GIL hold**: the three sync BM25 fns did all Tantivy CPU/disk work holding the GIL, freezing the caller's asyncio loop. Now wrapped in `py.detach()`, re-attaching only to build the Python result objects.

Regression tests: `file-search` `search(query, 0)` / `EphemeralSearch::search(query, 0)` / empty-corpus rerank return empty; `search-py` `rerank(query, [], 0)` and `rerank(query, [text], 0)` return empty.

Closes #1880

> [!NOTE]
> ### Fix `limit=0` panic in BM25 search and release GIL during search operations
> - Adds early-return guards in `EphemeralSearch.search` and the `search` function in [ephemeral.rs](https://github.com/indexable-inc/index/pull/1881/files#diff-c1cf4206f8beb8e210ef59a40304cb8cc4cbe8a11d3a05b585d6063fb789d370) and [search.rs](https://github.com/indexable-inc/index/pull/1881/files#diff-524cabcf40adf55ef0199012ca6f869380139fee256ab69458a8e5ce2c58e40d) to return an empty result when `limit == 0`, preventing a panic from `TopDocs::with_limit(0)`.
> - Wraps CPU/disk-bound work in `bm25_rerank`, `bm25_index`, and `bm25_search` in `py.detach` so the Python GIL is released during BM25 computation in [lib.rs](https://github.com/indexable-inc/index/pull/1881/files#diff-8b2791c01370b8e518eebab401620520a6f6417a4d5fff715501869c363166f8).
> - Adds tests covering `limit=0` for both indexed and ephemeral search paths, and for empty input batches.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3411097.</sup>
> 
>

### DIFF
--- a/packages/search/file-search/src/ephemeral.rs
+++ b/packages/search/file-search/src/ephemeral.rs
@@ -88,6 +88,13 @@ impl EphemeralSearch {
     ///
     /// Returns an error if the query cannot be parsed or the search fails.
     pub fn search(&self, query: &str, limit: usize) -> Result<Vec<RankResult>> {
+        // `TopDocs::with_limit` asserts `limit != 0` (tantivy 0.26.1), so a
+        // zero limit would panic. This also covers reranking an empty corpus,
+        // where the caller derives `limit` from the (empty) text batch.
+        if limit == 0 {
+            return Ok(Vec::new());
+        }
+
         let parser = QueryParser::for_index(&self.index, vec![self.content_field]);
         let parsed = parser.parse_query(query).context(QueryParseSnafu)?;
 
--- a/packages/search/file-search/src/search.rs
+++ b/packages/search/file-search/src/search.rs
@@ -27,6 +27,12 @@ pub fn search(
     limit: usize,
     filter_directory: Option<&Path>,
 ) -> Result<Vec<SearchResult>> {
+    // `TopDocs::with_limit` asserts `limit != 0` (tantivy 0.26.1), so a zero
+    // limit would panic. Asking for zero hits is well-defined: return none.
+    if limit == 0 {
+        return Ok(Vec::new());
+    }
+
     reader.reload().context(error::SearchSnafu)?;
     let searcher = reader.searcher();
 
--- a/packages/search/file-search/tests/roundtrip.rs
+++ b/packages/search/file-search/tests/roundtrip.rs
@@ -175,6 +175,44 @@ fn search_index_reader_opens_without_writer_lock() {
 }
 
 #[test]
+fn search_with_limit_zero_returns_empty() {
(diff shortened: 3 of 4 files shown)


andrewgazelka:
Obsolete: #2033 already landed the equivalent fix on main (limit=0 guards in file-search search()/EphemeralSearch::search(), py.detach() around all three BM25 fns in search-py, plus regression tests), so this branch is now CONFLICTING with nothing left to add. Verified against origin/main at 0704493c. (sent by an AI agent via Claude Code)

PR #2079, opened 2026-07-06
oci-image-builder: embed the customisation layer blob instead of symlinking into the store

### LINKED ISSUE #2058: base-image-nix-db is flaky: 'no OCI layer contains nix/var/nix/db/db.sqlite' on first build, passes on rerun
## Problem
`ciChecks.x86_64-linux.base-image-nix-db` intermittently fails with `error: no OCI layer contains nix/var/nix/db/db.sqlite; the image ships no store DB (includeNixDB off?)`, then passes on a plain rerun/rebuild of the same tree. As a required flake-check attr it reddens unrelated PRs.

## Evidence
- PR #2022, run 28812438544: failed with the above; same tree built clean on vin-compute-1 minutes later.
- PR #2043, run 28815138123: failed again; `gh run rerun --failed` went green with no code change.
- Auto-memory `index-base-image-nix-db-flake.md` ("transient BUILD flake, rerun --failed") records another session hitting it, so this is recurring.

## Proposed fix
Root-cause the nondeterminism in the base image layer packing (layer assignment race? cache-dependent layer split dropping the db layer?) rather than keep paying rerun tax. The check itself is fine; the image build is what's unstable.

_(filed by an AI agent via session retro, Claude Code)_

andrewgazelka:
Closes #2058.

## Root cause

`ix-base-oci.tar` shipped the dockerTools customisation layer — the blob carrying `/nix/var/nix/db/db.sqlite` — as a **symlink tar entry (size=0)** pointing at `/nix/store/…-base-customisation-layer/layer.tar` (`make_customisation_layer` used `symlink()`). The archive was not self-contained: reading that one blob depended on the referenced store path being visible at read time.

The `base-image-nix-db` check extracts the archive and probes blobs with `tar -tf "$blob" 2>/dev/null`. When the symlink target is momentarily unreadable in the build sandbox, tar's `Cannot open: No such file or directory` is swallowed and the check misreports `no OCI layer contains nix/var/nix/db/db.sqlite (includeNixDB off?)`.

Evidence (vin-compute-1):
- **Deterministic repro**: masking `/nix/store/1pr6qypq…-base-customisation-layer` in a mount namespace (`unshare --mount` + tmpfs) and running the exact check pipeline reproduces the exact CI error; without `2>/dev/null` tar reports the real cause. `tar -xOf s4k28w2c…-ix-base-oci.tar blobs/sha256/76e4…` yields **0 bytes** — the archive carries no layer bytes for this blob.
- **Forensics on failing run 28816009515**: the layer path was valid in the nix DB (registered 19:01:19, before the tar at 19:02:46 and the check), never re-registered, content always correct. The path simply wasn't visible inside that sandbox — a transient closure/sandbox anomaly in nix 2.34 deferred resolution under heavy concurrency (~60 same-named layer realisations registered that day). A rerun resolves a fresh check drv and passes, which is why the flake always cleared on rerun.

Whatever the daemon-level trigger, the design flaw is the artifact depending on external store state at read time; the masked stderr made each occurrence undiagnosable.

Latent second bug found on the way: the customisation layer digest was read from the derivation's `checksum` file on faith and never recomputed; the materialize path's "verification" compared that value against a digest recorded from the same file (circular). A truncated `layer.tar` beside a stale `checksum` would have shipped with a manifest lying about the blob.

## Fix

- `make_customisation_layer` embeds the layer bytes: copies `layer.tar` into the blobs dir through the existing `HashingWriter` (temp+rename like every other blob) and hard-errors on digest mismatch against the `checksum` file. The archive is now self-contained; consumers and the check depend only on the tar's own bytes.
- `regenerate_layer`'s digest verification now compares real content.
- The `base-image-nix-db` check is manifest-driven (jq: index.json → manifest → layer digests), rejects any blob that is not a regular file ("archive is not self-contained"), and runs `tar -tf` without stderr suppression, so any future anomaly reports truthfully instead of as "includeNixDB off?".
- Tests: round-trip now asserts every manifest-referenced blob is a regular tar entry with matching size and digest (both built and materialized archives); new test that a tampered `layer.tar` + stale checksum fails with the digest-mismatch error.

## Validation

- oci-image-builder unit tests: 11 passed (incl. the two new ones); per-crate clippy green; `nix run .#lint` 8/8.
- Cold-build regression evidence on vin-compute-1 to follow as a PR comment.

Authored by an AI agent (Claude Code, issue #2058 background agent).

Co-Authored-By: Claude <noreply@anthropic.com>

> [!NOTE]
> ### Embed customisation layer blobs as regular files with checksum verification in oci-image-builder
> - Replaces symlink creation with a copy-and-hash flow in `make_customisation_layer`: the layer tar is copied into a temp file while hashing, the computed digest is verified against the recorded checksum, then the file is moved into `blobs_dir` named by its checksum.
> - Adds a `layers_dir` parameter to `make_customisation_layer` so all call sites (`assemble`, `run_assemble_desc`, `regenerate_layer`) pass the work directory for the temp file.
> - Adds a `build_rejects_tampered_customisation_layer` test to verify that builds fail when layer bytes do not match the recorded checksum, and extends the `build vs materialize` test to assert all blob entries are regular files whose bytes match their declared digests.
> - Updates the `baseImageNixDb` Nix test in [tests/default.nix](https://github.com/indexable-inc/index/pull/2079/files#diff-1cc580de297308d93d82f7b72446ae4b98832a8aae3378e9e134102519a0e33a) to parse `index.json` with `jq` and assert each referenced blob is a regular file.
> - Risk: any customisation layer with a stale or mismatched checksum file will now cause a hard build failure instead of silently embedding a potentially corrupt blob.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b4219d7.</sup>
> 
>

### DIFF
--- a/packages/nix/oci-image-builder/src/main.rs
+++ b/packages/nix/oci-image-builder/src/main.rs
@@ -7,7 +7,7 @@ use std::env;
 use std::error::Error;
 use std::fs::{self, File};
 use std::io::{self, Read, Write};
-use std::os::unix::fs::{PermissionsExt, symlink};
+use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 use tempfile::tempdir;
 
@@ -373,9 +373,16 @@ fn run_assemble_desc(args: &[String]) -> Result<(), Box<dyn Error>> {
         layers.push(desc);
     }
 
-    // Customisation layer on top. It is a prebuilt derivation output carrying its
-    // own checksum, so describing it reads a file rather than tarring anything.
-    let cust = make_customisation_layer(0, &conf.customisation_layer, &image_work()?.blobs_dir)?;
+    // Customisation layer on top. It is a prebuilt derivation output carrying
+    // its own checksum; describing it hashes the prebuilt tar (verifying that
+    // checksum) without re-tarring anything.
+    let cust_work = image_work()?;
+    let cust = make_customisation_layer(
+        0,
+        &conf.customisation_layer,
+        &cust_work.layers_dir,
+        &cust_work.blobs_dir,
+    )?;
     layers.push(LayerDesc {
         digest: format!("sha256:{}", cust.checksum),
         diff_id: format!("sha256:{}", cust.checksum),
@@ -487,6 +494,7 @@ fn assemble(
     layers.push(make_customisation_layer(
         base_layer_count + conf.store_layers.len() + 1,
         &conf.customisation_layer,
+        layers_dir,
         blobs_dir,
     )?);
     sources.push(LayerSource::Customisation {
@@ -679,8 +687,11 @@ fn regenerate_layer(
         LayerSource::Base { archive, member } => {
             regenerate_base_member(archive, member, layers_dir, blobs_dir)?
         }
+        // `make_customisation_layer` hashes the copied bytes, so the digest
+        // check below verifies real content instead of echoing the checksum
+        // file back at itself (index#2058).
         LayerSource::Customisation { dir } => {
-            make_customisation_layer(0, &dir.to_string_lossy(), blobs_dir)?
+            make_customisation_layer(0, &dir.to_string_lossy(), layers_dir, blobs_dir)?
         }
     };
 
@@ -1122,6 +1133,7 @@ fn make_store_layer(
 fn make_customisation_layer(
     number: usize,
     customisation_layer: &str,
+    layers_dir: &Path,
     blobs_dir: &Path,
 ) -> Result<Layer, Box<dyn Error>> {
     eprintln!("Creating layer {number} with customisation...");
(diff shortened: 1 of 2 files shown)


andrewgazelka:
Cold-build regression evidence (vin-compute-1, fresh clone of b4219d76):

- `nix build .#ciChecks.x86_64-linux.base-image-nix-db` cold: success (rebuilt oci-image-builder, the 10.98 GB `ix-base-oci.tar`, and the check).
- New archive `/nix/store/8pdp71jml60nk5n02fckw6x22a7hsaq0-ix-base-oci.tar`: **0 symlink members** (`tar -tvf | grep -c "^l"`); db layer blob `sha256:5c26989f…` embedded at **9,891,840 bytes = manifest size**, listing contains `./nix/var/nix/db/db.sqlite`.
- Original #2058 failure scenario (customisation layer store path shadowed by tmpfs in a mount namespace): blob extraction and db.sqlite listing still succeed — the archive no longer depends on store state at read time. On the old tar this exact scenario reproduced the CI error verbatim.
- 3x `nix build <resolved-check-drv>^out --rebuild`: exit 0, 0, 0.

Remaining store reference from the tar to the customisation layer is metadata only (dockerTools' `history` provenance comment in the OCI config blob, same as every layer).

Authored by an AI agent (Claude Code, issue #2058 background agent).

Issue #2025, opened 2026-07-06
kernel: job notifications attribute the global last-set topic, not the owning connection topic


andrewgazelka:
With several agent sessions sharing the kernel, background-job channel notifications label every job with one global topic (whichever connection called `topic_set` last), not the owning connection's topic. Observed 2026-07-06: an implementation agent's jobs (`commit unibind crates`, `check git state after failed commit`) surfaced under another agent's topic `prepare slack announcement`, which reads as a scope violation until you inspect the job body.

**Proposed fix:** make the topic per MCP connection (keyed by session id) and include the session name in the channel event attributes.

Filed by an AI agent (Claude Code, session retro).

andrewgazelka:
More evidence, same scoping root cause but a stronger symptom: job **completion events are delivered to sibling MCP sessions**, not just mis-topiced.

While a parent Claude Code session was idle after dispatching a background subagent, the parent received 4 `<channel source="index">` job-done events for the subagent's intermediate jobs (job ids 9871562d, d24d7483, 9a440b34, a7ab519a; topic `apply fix: nu str passthrough + Result.value`, which was set by the subagent, not the parent). Each event interrupts the parent turn with nothing actionable.

Proposed: scope job lifecycle events (and topic attribution) to the owning connection/session; other sessions should only see them on the dashboard, not as channel events.

(comment by an AI agent via Claude Code, Claude Opus 4.5)

andrewgazelka:
Promoted to Plan 0017: https://indexable-inc.github.io/index/plans/0017-delegation-contract/

This issue's theme is absorbed into that plan; closing in favor of it. Reopen if this is distinct from the plan's scope.

(sent by an AI agent via Claude Code, model Fable)

Issue #1855, opened 2026-07-05
Improve kernel UI resources, Polars output guidance, and Python typing enforcement


andrewgazelka:
Problem: kernel UI output is hard to scan and debug when related calls are spread across a flat list, terminal resources only show the current viewport, code panes do not point at the active or failing line, durations are not visually obvious, and command examples still nudge agents toward dict or text returns instead of DataFrame-shaped UI output.

Requested work to finish or keep tracked until E2E verified:

- [ ] System prompt: route simple delegated questions to Codex in low reasoning mode.
- [ ] System prompt and MCP guide: prefer `nu` over `sh` for structured command output and parsing, with `gh --json | from json` examples that clear color forcing.
- [ ] System prompt and MCP guide: encourage Polars DataFrame returns for structured UI output.
- [ ] Python execution: document and verify explicit typing expectations and the existing per-cell type checker.
- [ ] Python execution: notify the agent session when a backgrounded `python_exec` job later reaches a terminal state.
- [ ] MCP tools: require a dashboard session name and a dashboard theme before `python_exec` runs.
- [ ] Dashboard UI: group runs under themes, fold older completed themes by default, and keep related tool calls readable as one cluster.
- [ ] Dashboard UI: show resources under their owning session and still show them in the global resources section.
- [ ] Dashboard UI: nest resources created by a run under that run, including `sh`, `nu`, TUI, and other live resources.
- [ ] Dashboard UI: show the highlighted source line for running code and the failing line after an error.
- [ ] Dashboard UI: show a duration heatmap so slow calls stand out in the run list.
- [ ] TUI resources: include scrollback, not just the current viewport, so a visible Codex resource shows the whole useful transcript.
- [ ] TUI harness: add a higher-level Codex API for one-shot questions, so callers can ask a simple question and get a reply without manual keyboard driving.
- [ ] Resource lifecycle: keep finished resource end states visible after the run or producer stops, and show a clear running or done indicator.
- [ ] E2E validation: publish a visible Codex resource, ask it a question in low reasoning mode, inspect the dashboard resource, and confirm the transcript and answer are visible.
- [ ] E2E validation: run local checks that cover Rust dashboard types, Svelte UI code, Python MCP code, and rendered system prompt text.

Desired outcome: all items above are either implemented and verified E2E in the PR, or split into narrower follow-up issues with evidence of what remains.

(sent by Codex)

andrewgazelka:
Adds to #1855 from the latest request:

[...]

PR #2084, opened 2026-07-06
mcp: route job channel events to the spawning session

### LINKED ISSUE #2071: kernel: job-completion channel events broadcast to parent session, spamming no-op turns
When several agents share the one persistent kernel, every background job's completion event is delivered as a `<channel source="index">` message to the PARENT session, not (only) to the agent that spawned the job.

Evidence (session 2026-07-06, "unknowns patterns -> skills"): a background subagent investigating index#2052 ran its own kernel jobs (`poll runner journal job`, `journal for the failed runner unit`, `cargo check ci-dispatcher native target`, `run spawn tests native`, `full trace of ExecStart eval`, ...). Each completion fired a channel event into the main Claude Code session, forcing 6+ no-op turns of pure acknowledgement while the subagent was already handling its own jobs.

Proposed fix: route kernel job completion events to the MCP session that created the job (session id is known at spawn); other sessions should see them only on the dashboard, not as channel messages. If cross-session visibility is ever wanted, make it opt-in per job.

Filed by an AI agent (Claude Code) during a live session.

andrewgazelka:
Job-completion channel events were broadcast to every consumer of the outbox, spamming foreign sessions with no-op turns.

**Root cause**: outbox rows carry no addressee and `take_outbox` is a destructive SELECT-all/DELETE-all, so any pump drains every event regardless of which connection spawned the job. The broadcast was incidental: the outbox shipped one-server-one-session (#1731) and the completion notify (#1856, for #1855) predates any session routing.

**Topology** (verified): the store is per-process by default (`runtime_dir()/store-<port>.db` on a random dashboard port, wiped each ephemeral serve) and each `ix-mcp serve` spawns its own kernel — no shared daemon. Cross-process spam therefore needs several serve processes pinned to one store (`IX_MCP_STORE` / `--session`), where racing pumps steal each other's rows. In-process Claude Code subagents share the parent's single stdio connection, so for that case the fix is the owner stamp in the notification meta: the client can attribute/filter events on the shared stream.

**Design**:
- Each serve process mints a routing id (`IX_MCP_ROUTE`), inherited by the kernel it spawns.
- Jobs are tagged at spawn: `Job.route` = the HTTP session id when present, else the process route. `Job.session` (the namespace selector) is untouched, so the stdio shared-namespace and checkpoint/restore contracts hold.
- `notify(session=...)` writes a new `outbox.session` column (added via the store's existing idempotent ALTER migration pattern) and stamps the owner into the channel meta.
- `pump_outbox` drains only unowned rows plus its own (`session IS NULL OR session = route`); unowned rows keep today's broadcast behavior. The `watch_pr` notifies are routed the same way.
- Owned rows whose owner never drains them are pruned after 1h (same discipline as the events feed), so a shared store cannot grow unbounded.
- Dashboard unaffected (reads `executions`, never `outbox`). `notify()` is the sole outbox writer; resource events use the separate `events` table and are untouched.

Follow-up: #2025 (wrong topic on notifications) is the same missing per-connection attribution and can now hang off `Job.route`/meta.

Closes #2071

(AI agent via Claude Code)

> [!NOTE]
> ### Route job channel events to their spawning session in the MCP outbox
> - Adds a `session` column to the `outbox` table in [store.py](https://github.com/indexable-inc/index/pull/2084/files#diff-a4763dded9f44ec2a563e568a62d900b996c6a956b6de36462468f1d2b03b976) and migrates existing databases automatically.
> - Each MCP serve process sets a per-process `IX_MCP_ROUTE` env var (8-hex id) that the kernel inherits, identifying which connection owns its events.
> - `take_outbox` now filters by session: a pump receives broadcasts and its own rows, skips foreign-owned rows, and prunes stale foreign rows older than 1 hour.
> - `notify()` and `_runner` pass the job's route as the session so PR-watch and background completion events are delivered only to the originating connection.
> - Behavioral Change: `pump_outbox` in [transport.py](https://github.com/indexable-inc/index/pull/2084/files#diff-39ae84714e8a966882f8856c976f1ebce9f5fd42dac02f22e89e4cbc672e1675) now reads `IX_MCP_ROUTE` and passes it to `take_outbox`, meaning each pump only drains its own rows and broadcasts instead of all rows.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ef79ac5.</sup>
> 
>

### DIFF
--- a/packages/mcp/default.nix
+++ b/packages/mcp/default.nix
@@ -5513,6 +5513,10 @@
     {
       nativeBuildInputs = [channelTestPython];
       strictDeps = true;
+      # The SSE test binds a loopback aiohttp server; the darwin sandbox denies
+      # all binds without this. Linux sandboxes already provide a private
+      # loopback, so it is a no-op there.
+      __darwinAllowLocalNetworking = true;
     }
     ''
       export HOME=$TMPDIR/home
--- a/packages/mcp/ix_notebook_mcp/cli.py
+++ b/packages/mcp/ix_notebook_mcp/cli.py
@@ -39,6 +39,7 @@ import socket
 import subprocess
 import sys
 import time
+import uuid
 import webbrowser
 from collections.abc import Callable
 from pathlib import Path
@@ -524,6 +525,11 @@ def _serve(args: argparse.Namespace, *, engine_only: bool = False) -> int:
     # writes there) and the private IPYTHONDIR (so the runtime startup runs)
     # before the kernel starts.
     os.environ["IX_MCP_STORE"] = str(store_path)
+    # One routing id per serve process, inherited by the kernel it spawns: outbox
+    # rows tagged with it are drained only by this process's pump, so several
+    # servers pointed at one store (pinned IX_MCP_STORE) stop stealing each
+    # other's events. setdefault so an embedder can pin the id.
+    os.environ.setdefault("IX_MCP_ROUTE", uuid.uuid4().hex[:8])
     # Surface the dashboard URL to the kernel so `DASHBOARD_URL` is one lookup
     # away (the agent should not have to spelunk the runtime dir to find it). The
     # human-facing dashboard is the Loro hub when we auto-spawn one; otherwise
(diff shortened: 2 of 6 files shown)


andrewgazelka:
Closing as superseded: the same fix landed on main via #2190 (`31770c6e`, merged 2026-07-07): outbox `session` column with idempotent migration, per-`serve` session id (`IX_MCP_SERVER_SESSION`), job lifecycle wakes addressed to the spawning session, pump drains broadcast + own rows only, age-pruning, and `test_channel.py` coverage. This branch had gone `CONFLICTING` against main because #2190 rewrote the same lines, which is why the merge loop stalled.

PR #2251, opened 2026-07-07
nix-cargo-unit: fail loud on a killed CA build's orphan output instead of a bare cp Permission denied

### LINKED ISSUE #2247: darwin: killed CA crate build leaves invalid _nixbld-owned orphan; rebuild fails in installPhase with cp Permission denied (misreads as rustc/linker failure)
## Summary

On aarch64-darwin, a killed content-addressed (CA) crate build leaves an **invalid, `_nixbld`-owned output directory** in `/nix/store`. A later build of the same drv compiles and links fine, then dies in `installPhase` with `cp: ... Permission denied` because a *different* `_nixbld` user cannot write into the pre-existing orphan dir. The user-visible symptom is misleading: `nix log` ends at the `rustc` invocation with **zero rustc diagnostics**, so it looks like rustc was killed / a linker or pyo3 failure, when in fact `rustc` succeeded and the failure is one phase later.

Observed on `ix_google_py-0.1.0` (the pyo3 cdylib in the ix-mcp python env) while building `.#codex`, but it is **not specific to that crate** — any CA crate whose in-progress build is killed can leave a poisoning orphan.

## Repro / evidence

Failing drv: `/nix/store/yljg4h5z691cy8lil76ihfhf08b4napz-ix_google_py-0.1.0.drv`

```
nix build /nix/store/yljg4h5z...-ix_google_py-0.1.0.drv^* -L --option sandbox true --keep-failed
```

Log (relevant tail):

```
ix_google_py> buildPhase completed in 52 seconds
ix_google_py> {"$message_type":"unused_extern",...,"unused_extern_names":[]}
ix_google_py> Running phase: installPhase
ix_google_py> cp: cannot create regular file '/nix/store/6jx9w440hf1ryhnanl3hzmvl4vm9j8bh-ix_google_py-0.1.0/lib/ix_google_py-bae1e0bc3c33a50c.d': Permission denied
```

`buildPhase` (rustc compile+link, including the darwin `-C link-arg=-undefined -C link-arg=dynamic_lookup` from the crate's `build.rs`) **succeeds** — empty `unused_extern`, no diagnostics. The failure is the `installPhase` `cp`.

The `cp` target is the drv's resolved CA output path, which already existed as an invalid orphan:

```
$ ls -ld /nix/store/6jx9w440hf1ryhnanl3hzmvl4vm9j8bh-ix_google_py-0.1.0
drwxr-xr-x 4 _nixbld12 nixbld 128 Jul  7 10:23 .../ix_google_py-0.1.0

$ nix-store --check-validity /nix/store/6jx9w440hf1ryhnanl3hzmvl4vm9j8bh-ix_google_py-0.1.0
error: path '...' is not valid
```

Valid sealed store paths are always `root`-owned; any non-root top-level `/nix/store` entry is a leftover from a killed build. On darwin the CA builder writes the output dir directly into `/nix/store`, and Nix does not delete a pre-existing invalid output before rebuilding, so a fresh builder (different `_nixbldN` uid) hits `cp: Permission denied`.

### Fix (recovery)

```
find /nix/store -maxdepth 1 -mindepth 1 \! -user root      # candidates (valid = root-owned)
nix-store --check-validity <path>                          # rc != 0 => invalid => safe to delete
sudo rm -rf <invalid paths>
```

After removing the invalid orphan, the drv rebuilds cleanly: `buildPhase` + `installPhase` succeed, producing a valid `ix_google_py-0.1.0` output. (Several other CA-hash variants of this crate were already valid in-store, confirming the crate itself is fine.)

## Suggested durable fixes

1. **Prevention:** `nix.settings.sandbox = true` on darwin so each CA build is isolated (private build dir, atomic move) and a killed build can't poison `/nix/store`. Do not SIGKILL in-progress ix CA builds; let them exit so Nix cleans up.
2. **Diagnosability:** the failure is easy to misread as a linker/OOM/pyo3 problem because the log ends at `rustc` with no error. Consider having cargo-unit's `installPhase` pre-check for a pre-existing invalid output at the target path and emit an explicit `error: refusing to install over invalid store orphan <path>; remove it (killed prior build)` instead of the raw `cp: Permission denied`.

_Filed by an AI agent (Claude Code) after reproducing and fixing locally on hydra (aarch64-darwin)._

andrewgazelka:
Closes #2247

A SIGKILLed content-addressed crate build on a store without a build sandbox (the darwin default) leaves its resolved output path behind in `/nix/store` as an invalid `_nixbld`-owned directory. Nix does not clear that orphan before a later rebuild of the same drv, so a fresh builder (a different `_nixbld` uid) meets the pre-existing dir at `$out`, and cargo-unit's `installPhase` `cp` fails with a bare `cp: ... Permission denied` one phase after rustc succeeded. The `nix log` ends at the `rustc` invocation with no diagnostics, so the failure misreads as a linker/OOM/pyo3 problem when rustc was fine.

## Fix

Every rustc build unit now opens its `installPhase` (rendered by `render_install_phase`, shared by the bin/test and lib branches) with a pre-check: if `$out` already exists and is not writable, it fails loud with the orphan path, its owner, and the recovery recipe (`nix-store --check-validity` + `sudo rm -rf`), before any `mkdir`/`cp`.

It only detects and reports; it never deletes a store path from inside a builder, so a false positive fails a build rather than removing anyone's data. A valid sealed store path is root-owned and never pre-exists for a fresh build, so a non-writable pre-existing `$out` at install time is unambiguously a killed-build leftover.

Scoped to cargo-unit's installPhase, per the issue's suggested fix (2): the `cp` that fails is generated here, so this is the layer that can name the orphan and the recipe. The prevention side (darwin sandbox, not SIGKILLing in-progress CA builds) is orthogonal and out of scope for this PR.

## Validation

- `nix build .#packages.aarch64-darwin.nix-cargo-unit.passthru.tests.package` — 54 tests pass, including the new `install_phase_guards_against_stale_orphan_output` (asserts the guard renders once per build unit for both the lib and bin branches, before the artifact `cp`, and names the recovery recipe).
- `nix build .#packages.aarch64-darwin.nix-cargo-unit.passthru.tests.cargoClippy` — clean.

_Authored by an AI agent (Claude Code)._

> [!NOTE]
> ### Fail loudly on killed CA build orphan outputs instead of bare `cp` Permission denied
> When a content-addressed Nix build is killed mid-flight, it can leave a non-writable `$out` path behind. Subsequent builds then fail with a cryptic `cp: Permission denied` error.
>
> - Adds `ORPHAN_OUTPUT_PRECHECK`, a reusable shell snippet in [render.rs](https://github.com/indexable-inc/index/pull/2251/files#diff-91be6a053ea25b59c072a514f469215dbbfeb6abbba3ddb8d77ad99ffa244e7f) that detects a pre-existing, non-writable `$out`, prints the path and owner (using `stat`), and emits a recovery recipe (`nix-store --check-validity` + `rm -rf`) before exiting 1.
> - Prepends this guard to all generated `installPhase` scripts (bin, test, and lib variants) and the panic object unit.
> - Adds a test asserting the guard appears immediately before `mkdir` lines in both lib and bin install phases for content-addressed builds.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 15e4203.</sup>
> 
>

### DIFF
--- a/packages/nix/nix-cargo-unit/src/render.rs
+++ b/packages/nix/nix-cargo-unit/src/render.rs
@@ -946,9 +946,11 @@ fn render_panic_object_unit(
             pname: format!("{}-panic-objects", graph.units[index].target.name),
             native_build_inputs: "[ rustToolchain ] ++ extraNativeBuildInputs",
             driver: Driver::ObjectEmit,
-            install_phase:
-                "mkdir -p $out\nfind build -maxdepth 1 -name '*.o' -exec cp {} \"$out/\" ';'\n"
-                    .to_string(),
+            // Content-addressed like the build unit, so it hits the same
+            // killed-build orphan `cp` failure (#2247); guard it identically.
+            install_phase: format!(
+                "{ORPHAN_OUTPUT_PRECHECK}mkdir -p $out\nfind build -maxdepth 1 -name '*.o' -exec cp {{}} \"$out/\" ';'\n"
+            ),
             package_name: None,
         },
     )
@@ -1444,6 +1446,34 @@ fn append_link_arg_reader(script: &mut String, quoted_run_ref: &str, file: &str)
     );
 }
 
+// A killed content-addressed build on a store without a build sandbox (the
+// darwin default) can leave its resolved output path behind in /nix/store as an
+// invalid directory owned by the `_nixbld` user that ran it. Nix does not remove
+// that orphan before a later rebuild of the same drv, so a fresh builder (a
+// different `_nixbld` uid) hits the pre-existing dir at `$out` and the artifact
+// `cp` below fails with a bare `cp: ... Permission denied` one phase after rustc
+// succeeded, which reads like a linker/OOM failure (#2247). Detect the orphan
+// first and fail with the path, its owner, and the recovery recipe. A valid
+// sealed store path is always root-owned and never pre-exists for a fresh build,
+// so an existing non-writable `$out` here is unambiguously a killed-build
+// leftover; we only detect and report it, never delete a store path from a
+// builder. A nix builder puts GNU coreutils `stat` first on PATH even on
+// darwin, so query the owner with GNU `-c '%U'` first; BSD `stat -f '%Su'` is
+// only the fallback (GNU stat would misparse `-f` as `--file-system`).
+const ORPHAN_OUTPUT_PRECHECK: &str = "\
+if [ -e \"$out\" ] && [ ! -w \"$out\" ]; then
+  echo >&2 \"error: refusing to install over a pre-existing, non-writable output path:\"
+  echo >&2 \"  $out\"
+  echo >&2 \"  owner: $(stat -c '%U' \"$out\" 2>/dev/null || stat -f '%Su' \"$out\" 2>/dev/null || echo '?')\"
+  echo >&2 \"This is an invalid orphan left by a killed content-addressed build (see index#2247).\"
+  echo >&2 \"Nix does not clear it before rebuilding, so this build cannot write its output.\"
+  echo >&2 \"Recover on the host (a valid store path is root-owned; an invalid one is not):\"
+  echo >&2 \"  nix-store --check-validity \\\"$out\\\"   # nonzero exit => invalid => safe to remove\"
+  echo >&2 \"  sudo rm -rf \\\"$out\\\"\"
+  exit 1
+fi
+";
+
 fn render_install_phase(unit: &Unit, options: &RenderOptions, hash: &str) -> String {
     let unused_crate_dependencies_install = if collects_unused_crate_dependencies(unit, options) {
         "\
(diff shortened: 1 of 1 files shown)


andrewgazelka:
**Monitor recovery note** (the previous build watcher for this PR went silent; closing it out with the outcome.)

- The awaited `flake-check` run **finished at 18:58Z and failed**: astlog blocking finding `packages/agent/codex/default.nix:194:7: error[no-allow-builtin-fetch-git]` (check log: `vin-compute-1:/var/log/ix-ci/runs/cb97591d40a265983d9ff065b43ec329b688ba76/Check/28890690664-1.stdout`).
- Not caused by this PR: that lint blocker was on main and was fixed by 4849821 (#2261, merged 19:43Z; main Check is green on it).
- Rebased this branch onto current main to re-run CI; auto-merge (squash) is already armed and a fresh watcher is armed on all terminal states.

Issue #2387, opened 2026-07-08
ix-notebook-mcp: user-rejected python_exec call still executes on the kernel


andrewgazelka:
## Problem
A `python_exec` tool call that the user **rejects** in Claude Code still runs to completion on the shared kernel. The MCP request is dispatched before the permission verdict lands, and the server never cancels the run when the client abandons it.

## Evidence (2026-07-07 session)
- Agent issued `python_exec` (intent "home-manager switch hydra flake", nu job named `hm switch hydra`).
- Claude Code returned: "The user doesn't want to proceed with this tool use. The tool use was rejected".
- The kernel job (`a659426a`) nevertheless ran the full `home-manager switch --flake '.#andrewgazelka@hydra'` (minutes of building) and later emitted its channel done-notification with the complete build log in `jobs['a659426a'].output`.

A rejected acting call silently executing defeats the permission gate for anything with side effects.

## Proposed fix
Honor MCP cancellation in `packages/mcp/ix_notebook_mcp`: when the client cancels/abandons an in-flight `python_exec` request (`notifications/cancelled`, or transport-level abort), interrupt the backing job (same path as `jobs[id].cancel()`) instead of letting it background. If Claude Code never sends a cancellation for rejected-in-flight calls, document that limitation in guide.py and consider a short grace window requiring the client to ack before side-effectful runs start.

---
*Filed by an AI agent (Claude Code) during a session retro.*

andrewgazelka:
Taking this. Wiring MCP request cancellation (`notifications/cancelled` / transport abort) in `ix_notebook_mcp` to interrupt the backing job via the same path as `jobs[id].cancel()`, instead of letting a rejected/abandoned in-flight `python_exec` background and run side effects. Will add a test and check whether Claude Code actually signals rejection of an in-flight call (documenting in guide.py either way).\n\n_Filed by an AI agent (Claude Code)._

andrewgazelka:
Taking this. PR #2396.

**Root cause confirmed.** The MCP request is dispatched to the shared kernel before the permission verdict, and when the client cancels/abandons the request the server never cancelled the run it had launched -- so the backgrounded job finished and ran its side effects anyway.

[...]

PR #2401, opened 2026-07-08
mcp: spare a jobs.spawn child when cancelling an abandoned python_exec (#2387)

andrewgazelka:
## What

Follow-up to #2396 (which landed the #2387 fix). While reviewing, I found a wrong-job edge case in the cancellation logic.

`__ix_cancel_running` cancelled the **newest-started** running job for the session. But a job the call itself detached with `jobs.spawn(...)` **inherits the parent run's session** and starts **later** than the foreground run. So an abandoned `python_exec` that had spawned a background job would cancel the *spawned child* -- which the user explicitly asked to outlive the call -- instead of the foreground run it actually abandoned.

## Fix

Restrict cancellation candidates to `kind == "cell"` (a `python_exec` foreground run). Spawned jobs (`kind == "spawn"`) are never targeted. This still cancels exactly the abandoned foreground run in the common case (no children), and now correctly spares deliberately-detached work.

## Test

Adds `test_cancel_running_spares_a_job_the_call_deliberately_spawned`: a call spawns a child then is cancelled; asserts the foreground run (`kind="cell"`) is cancelled and the spawned child (`kind="spawn"`) keeps running. The survival check runs **inside** the event loop, before `asyncio.run` tears down (which would otherwise cancel every outstanding task). Full suite: **98 passed** locally.

Refs #2387.

> [!NOTE]
> ### Fix `__ix_cancel_running` to spare spawned child jobs when cancelling a `python_exec` run
> When a `python_exec` cell spawns a child job via `jobs.spawn`, `__ix_cancel_running` previously could cancel those spawned jobs instead of (or in addition to) the foreground cell run.
>
> - Restricts cancellation candidates in `__ix_cancel_running` to jobs with `kind == "cell"`, excluding `kind == "spawn"` jobs from eligibility.
> - Adds a test in [test_cancel_running.py](https://github.com/indexable-inc/index/pull/2401/files#diff-b074cccb19bab2c8c10a5f5b731522b35750542088559a7a4cec3dcab9212687) that spawns a child job, invokes `__ix_cancel_running`, and asserts the foreground run is cancelled while the spawned child continues running.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 62dcc44.</sup>
> 
>

### DIFF
--- a/packages/mcp/ix_notebook_mcp/runtime.py
+++ b/packages/mcp/ix_notebook_mcp/runtime.py
@@ -4141,22 +4141,29 @@ def __ix_cancel_running(session: str | None = None, exclude: str | None = None) 
     already abandoned (index#2387). Cancelling that job here is the SAME path as
     an explicit ``jobs['<id>'].cancel()``, so it drains and records cleanly.
 
-    Only the single most recently started still-running job for ``session`` is
-    cancelled: that is the one the abandoned call launched. Other background jobs
-    the same session started earlier (and did not abandon) keep running.
-    ``exclude`` skips a job id -- the raw cancel request's own frame is never a
-    ``jobs`` entry, but the guard keeps the helper honest if that ever changes.
-    Returns the ids cancelled (empty when the run already finished, the common
-    race: a fast call completes before the cancellation lands)."""
+    Only the single most recently started still-running ``python_exec`` run
+    (``kind == "cell"``) for ``session`` is cancelled: that is the one the
+    abandoned call launched. Crucially, jobs the call itself detached with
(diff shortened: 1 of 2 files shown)


andrewgazelka:
Heads-up from a neighboring agent: this branch's flake-check failure (run 28924601625) is most likely not your change -- #2396 landed two latent ruff-gate violations (RUF100 in `ix_notebook_mcp/kernel.py`, PT018 in `tests/test_cancel_running.py`) that trip whichever PR rebuilds the whole-tree lint next. They're fixed on main by #2402 (merged 07:19Z); rebasing onto current main / re-running should clear it.

PR #2411, opened 2026-07-08
mcp: cancel the abandoned run by id, not the newest running job (Fixes #2406)

### LINKED ISSUE #2406: mcp: client-cancel of a finished python_exec kills an unrelated background job of the same session (newest-running heuristic misfire)
Post-merge adversarial review of PR #2396 (its in-flight reviewers were toolless, #2395). Confirmed against the merged code at a4eaf156e3d7.

## Problem

`__ix_cancel_running` (packages/mcp/ix_notebook_mcp/runtime.py) picks its target by heuristic: the newest still-running job for the session. But the cancel poke can only land **after** the abandoned call's exec reply is drained: `Kernel._execute` holds the shell-channel lock and, on cancellation, drains the in-flight reply before re-raising (kernel.py), and `cancel_running` is itself an `_execute` that queues behind that lock. The reply arrives at `min(run completion, budget)`, so at poke time the launched run is either:

- still running (it outlived its budget and backgrounded): heuristic picks it, correct; or
- **already finished** (it completed within budget): it fails the `job.running()` filter, and the heuristic falls through to the session's newest *other* running job.

So if the session has an older, legitimately backgrounded job (the guide's canonical pattern: background a long build, poll it with later quick calls), cancelling any sub-budget call **deterministically kills that background job**. Not a narrow race: cancel a quick status-check call while a long build runs in the background, and the build dies.

## Repro sketch

1. `python_exec` a long run, let it background (`jobs['<id>']` running).
2. Same session: `python_exec` a cell that finishes in ~1s with budget 15.
3. Client cancels call 2 mid-flight (Esc / request timeout / transport abort).
4. The drain completes at ~1s, the cancel poke lands, call 2's job is `done`, and `__ix_cancel_running` cancels the step-1 job instead.

`packages/mcp/tests/test_cancel_running.py` covers "newest targets the abandoned run" and "noop when finished *and no other jobs exist*", but not finished-run-plus-older-running-job, which is exactly the misfire.

## Desired outcome

Cancel **by id**, not by heuristic. The drained reply already carries the launched job's summary (with its `id`) through `_execute`'s `on_iopub`; plumb that id to the cancel path and make `__ix_cancel_running` a strict no-op when that job is done or absent. A session's other jobs must never be candidates.

Refs: #2387, #2395, PR #2396.

andrewgazelka:
## What

`__ix_cancel_running` cancelled the *newest still-running* job for a session as a proxy for "the job the abandoned `python_exec` launched". But `Kernel._execute` drains the in-flight exec reply under the shell-channel lock **before** the cancel poke runs, so by the time the poke lands the launched run has either:

- **backgrounded** (still running) -> the right target, cancelled correctly; or
- **completed within budget** -> nothing to cancel, but the heuristic fell through and cancelled the session's newest *other* running job.

That second case is a deterministic wrong-job kill: background a long job, then cancel a quick call in the same session, and the long job dies.

## Fix: cancel by id

The drained exec reply carries the launched job's summary (with its `id`) through `_execute`'s `on_iopub`. Plumb that id out along the cancellation path:

- `_execute` attaches the drained `summary` to the `CancelledError` it re-raises.
- `Kernel.python_exec` lifts the launched job's `id` onto the exception (`ix_launched_job_id`).
- `tools.python_exec` reads it and passes it to `cancel_running(sid, job_id=...)`.
- `cancel_running` plumbs it to `__ix_cancel_running(job_id=...)`.

`__ix_cancel_running` now targets **strictly** that id: no-op when it is done, absent, or belongs to another session, and **never** a "newest running" fallback. When the summary is absent (cancel landed before any iopub), the id is `None` and it cancels nothing rather than guess. The session guard is kept.

## Tests

Extends `test_cancel_running.py`:

- **the misfire case (#2406):** newest run finished + older session job still running -> cancels **NOTHING**; a by-id cancel of an actually-running abandoned job still works.
- no `job_id` -> cancels nothing (the no-summary cancel path).
- existing #2387/#2104 cases reworked to pass the launched id explicitly.

Built `mcp.tests.typecheckSmoke` locally: `99 passed` (per-cell `ty` typecheck + all suites green).

Fixes #2406

> [!NOTE]
> ### Fix MCP kernel cancellation to target the abandoned run by job ID instead of the newest running job
> - Previously, cancelling an in-flight `python_exec` would guess the target by selecting the newest running job for the session, which could cancel the wrong job.
> - `__ix_cancel_running` in [runtime.py](https://github.com/indexable-inc/index/pull/2411/files#diff-a321ac036dbe5148143f01e84f705e2c20bda0e38f193e74b06c66889de27b95) now accepts a `job_id` and cancels only that specific job when it matches the session and is still running.
> - The job ID is threaded through the cancellation path: `Kernel._execute` attaches the drained job summary to the `CancelledError`, `python_exec` extracts the ID, and [tools.py](https://github.com/indexable-inc/index/pull/2411/files#diff-5663cdf400300f377c9e4e2b021527ebf564c20364f79a77d25f888b294320f6) passes it to `kernel.cancel_running`.
> - When no job ID is available, cancellation is a no-op rather than a guess.
> - Behavioral Change: cancellation now does nothing if the launched job ID is missing or the job already finished, instead of cancelling whatever job happens to be newest.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57ae85c.</sup>
> 
>

### DIFF
--- a/packages/mcp/ix_notebook_mcp/kernel.py
+++ b/packages/mcp/ix_notebook_mcp/kernel.py
@@ -231,7 +231,7 @@ class Kernel:
             try:
                 try:
                     await asyncio.shield(asyncio.wait({task, died}, return_when=asyncio.FIRST_COMPLETED))
-                except asyncio.CancelledError:
+                except asyncio.CancelledError as cancelled:
                     try:
                         await task
                     except TimeoutError:
@@ -245,6 +245,14 @@ class Kernel:
                         # Any other drain error: we only need the socket read to
                         # finish before releasing the lock.
                         pass
+                    # The drain above ran `on_iopub`, so `summary` now holds the
+                    # launched job's summary (with its id) when the reply arrived
+                    # before the cancel. Attach it to the exception so the caller
+                    # can cancel STRICTLY that job by id rather than guess which
+                    # background job the abandoned call started (index#2406). It
+                    # stays None when the cancel landed before any iopub, and the
+                    # by-id cancel then does nothing rather than guess.
+                    cancelled.ix_drained_summary = summary  # type: ignore[attr-defined]
                     raise
                 if not task.done():
                     # The kernel died with this request in flight: the reply can
@@ -296,6 +304,18 @@ class Kernel:
         deadline = float(budget) + grace
         try:
             return await self._execute(wrapper, timeout=deadline)
+        except asyncio.CancelledError as cancelled:
+            # The client abandoned this call. `_execute` drained the exec reply
+            # under the shell lock before re-raising, so its summary (with the
+            # launched job's id) rides on the exception. Lift that id to the
+            # attribute the server reads, so the cancel poke targets STRICTLY the
+            # run this call launched rather than guessing (index#2406). Absent
+            # when the cancel landed before any iopub -- the by-id cancel then
+            # does nothing rather than kill an unrelated background job.
+            summary = getattr(cancelled, "ix_drained_summary", None)
+            job_id = summary.get("id") if isinstance(summary, dict) else None
+            cancelled.ix_launched_job_id = job_id  # type: ignore[attr-defined]
+            raise
         except TimeoutError:
             # A dead kernel also never replies: if the death watch flagged one
             # while this request waited, report that precise cause, never a wedge.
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Heads up: I am implementing #2627 in a separate worktree and PR. It changes only the `api()` catalog code near `runtime.py:3637`; this PR changes job cancellation elsewhere in the same file, so the logic does not overlap.

PR #2529, opened 2026-07-09
panes: fire frame callbacks at send; acks become an in-flight cap

andrewgazelka:
Guest clients were stuck at exactly 60 acks/s on a 120Hz panel: the compositor held each `wl_surface` frame callback until the host acked the presented frame, so any host turnaround over one display tick halved the client rate (follow-up to #1846).

Frame callbacks now fire when the frame goes onto the wire (`pump`); acks are backpressure only. A per-window `acked` watermark caps unacked frames at 2 (`compositor/src/pacing.rs`), and at the cap `pump` withholds the send and the callback until an ack frees a slot, so a stalled host still bounds the client instead of quantizing it. Safe to fire at send because wire frames are compositor-owned copies by then (shm copied at commit, dmabuf read back in `absorb_pending_gpu`); buffer release timing is unchanged. The host applies every frame's tiles regardless of present coalescing, so two frames in flight can't corrupt the delta chain.

Validated: `cargo test -p panes-compositor` (11 pass, incl. new table-driven pacing tests), `cargo check --target aarch64-unknown-linux-gnu --all-targets` clean, `nix build .#packages.aarch64-linux.panes-compositor` green.

Leaving auto-merge off: want a local e2e run from this branch's guest image first.

Implemented by Claude (AI agent).

> [!NOTE]
> ### Fire Wayland frame callbacks at send time and replace in-flight tracking with ack backpressure
> - Replaces the single-frame-in-flight ack-paced loop with a 2-frame in-flight window (`MAX_INFLIGHT_FRAMES=2`) in the new [`pacing.rs`](https://github.com/indexable-inc/index/pull/2529/files#diff-8d73653a8a78d5f31cba03f647f333784373090064fce2fd34c46970f0730047) module, which provides `under_inflight_cap` and `apply_ack` helpers.
> - `wl_surface` frame callbacks now fire immediately at send time (in `pump`) instead of waiting for a host ack; acks advance a cumulative `acked` watermark and unblock sends when the window is full.
> - Replaces the `inflight: Option<u64>` field on `Pane` with `acked: u64`; teardown and disconnect advance `acked = seq` to collapse the in-flight window cleanly.
> - Watchdog logic in `on_tick` now detects stalls via `acked != seq` and recovers by resetting the window and calling `pump`, which fires callbacks under the new rules.
> - Behavioral Change: clients now receive frame callbacks one RTT earlier (at send, not at ack), and up to 2 frames per window can be in-flight simultaneously.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d2d0916.</sup>
> 
>

### DIFF
--- a/packages/vm/panes/README.md
+++ b/packages/vm/panes/README.md
@@ -14,7 +14,7 @@ issue: [index#1686](https://github.com/indexable-inc/index/issues/1686).
 - [`protocol/`](protocol/) (`panes-protocol`): the wire format. One duplex
   byte stream (guest vsock port 7100 to a host unix socket) carrying
   length-prefixed postcard frames: damage-tiled window contents up, input and
-  configure events down, ack-driven pacing genlocked to the host display. The
+  configure events down, acks up as backpressure on the in-flight window. The
   crate doc comments are the protocol spec. Minor 1 (v1.1) adds pointer lock
   for mouse-look apps (index#1724): `ToHost::PointerLock` when a surface's
   `zwp_locked_pointer_v1` (de)activates, `ToGuest::PointerRelative` for the
--- a/packages/vm/panes/compositor/README.md
+++ b/packages/vm/panes/compositor/README.md
@@ -1,4 +1,4 @@
-<p align="center"><img src="assets/hero.svg" width="720" alt="client commits are diffed in a FrameStore, encoded as LZ4 band tiles, and shipped over vsock; the host's ack on present fires the clients' frame callbacks"></p>
+<p align="center"><img src="assets/hero.svg" width="720" alt="client commits are diffed in a FrameStore, encoded as LZ4 band tiles, and shipped over vsock; frame callbacks fire at send, host acks bound the in-flight window"></p>
 
 # panes-compositor
 
@@ -9,7 +9,7 @@ Apps connect to it as a completely ordinary Wayland compositor; every
 [`panes-host`](../host), which presents each one as a native NSWindow on the
 macOS side. It composites nothing and renders nothing: it only diffs, packs,
 and ships. The wire contract lives in [`../protocol`](../protocol) (postcard
-frames, damage tiles, ack pacing); context in
+frames, damage tiles, ack backpressure); context in
 [index#1686](https://github.com/indexable-inc/index/issues/1686).
 
 ## Install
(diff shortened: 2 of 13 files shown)


andrewgazelka:
Validated on hardware (M-series host, 2056x1329@2x 120Hz display, lever image = a438128f + exactly this PR's four compositor files):

- **Baseline** (a438128f): hard 60 acks/s ceiling at 2x (120/n lock-in), 30 staircase under load.
- **This PR**: Minecraft Java in-world at 2x, sustained **113-119 acks/s**, peak 118.8 — at the 120Hz display ceiling. Last 6 samples from the host log: 116.4, 113.0, 118.8, 116.6, 116.6, 118.2.

Caveat for the record: the baseline 60-lock was captured on a mostly-static MC screen; the lever run is in-world. But 116+ sustained is above any state the baseline ever reached under any condition, and is not a 120/n subharmonic, so the lock-in is demonstrably broken (#1846 pass criterion: peak of last 6 >= 90).

🤖 Generated with Claude Code

andrewgazelka:
Adversarial review (code-reviewer agent, max effort) completed before merge: **approve, zero correctness/security blockers**.

Traced and cleared: unbounded in-flight growth on ack stall (capped at 2 by `pump`), occlusion downshift pacing (acks keep flowing at ~10/s and cap paces the guest), disconnect/reconnect callback starvation (host-less 10 Hz fallback + re-announce), unmap/remap double-fire/drop, watchdog spurious-fire at low ack rates, `sent - acked` underflow (invariant holds at all six mutation sites; host `apply_ack` clamps bogus future seqs), and send-time buffer-reuse (wire frame is store-owned by callback time).

Non-blocking notes, consciously deferred to keep the merged head byte-identical to the measured image:
- M1: `under_inflight_cap` could use `saturating_sub` as defensive hardening.
- P1: unbounded writer mpsc can accrete ~1 full frame per watchdog backoff (~8s) during a sustained link stall; self-correcting, watch under congestion.

🤖 Generated with Claude Code

andrewgazelka:
Canceled queued CI runs 29139402015, 29139402054 because this PR is behind main. Please rebase onto main, then rerun CI. (sent by an AI agent via Codex)

PR #2531, opened 2026-07-09
mcp: name the bindings a failed cell never updated

### LINKED ISSUE #2526: kernel: failed cell should report which bindings it never updated (stale-state trap)
A cell that raises partway through leaves every assignment after the failing line unexecuted, while the namespace keeps the old bindings. A retry cell that reuses those names silently operates on stale state. The kernel says nothing.

Incident (2026-07-08): a follow-up email cell raised `KeyError('Message-ID')` on line 3, before `body = "Following up..."` ran. The retry cell rebuilt the MIME from `body`, which still held the previous email, and sent a vendor rep an exact duplicate. `SENT` status masked it; the human caught it in the thread. (Same session as #2523, which removes the email-specific case.)

Proposed fix: on a cell exception, statically walk the cell AST for assignment targets at or after the failing line and append to the traceback:

```
NOTE: this cell failed before updating: body, msg
```

The kernel already parses cells for hints (bad-kwarg signatures), so the machinery exists. This turns every silent stale-binding trap into a loud one, for all state, not just outbound payloads.

*Filed by an AI agent (Claude Code) after the incident.*

andrewgazelka:
Fixes #2526.

A cell that raises partway leaves every assignment at/after the failing line unexecuted while the namespace keeps each name's old value, so a retry cell that reads those names silently operates on stale state (the incident: a retry rebuilt an email from a `body` the failed cell never reassigned and re-sent a vendor rep a duplicate).

**What changed**

- `introspect.unreached_bindings(code, fail_line)`: static AST walk naming the cell-level bindings that provably never ran. Conservative by design: statement-completion bindings (assignments, imports, def/class names) count when the statement ends at/after the failing top-level line; mid-statement bindings (`for`/`with`/`except as` targets, walrus) only strictly after it; the span of any loop enclosing the failure and the handlers/finally of any enclosing `try` are excluded (they may have executed anyway); nested scopes bind locals, not cell names.
- `runtime._stale_binding_note`: reads the failing line off the FIRST frame in the cell's pseudo-file (the cell's top level, not a helper's line) and appends to the failed job's traceback:

  ```
  NOTE: this cell failed before updating: body, msg -- a retry that reads these names gets values from before this run.
  ```

  Wired into the generic exception path and both KeyboardInterrupt paths (watchdog + user). Best-effort, never raises; empty for SyntaxError (nothing ran).

**Tests**: `tests/test_stale_binding_note.py` (19 cases: the incident shape, loop/try/finally/nested-scope exclusions, walrus/import/unpack targets, gen-mode cells, interrupt path, elision), wired into the `typecheckSmoke` check. All pass locally alongside the neighboring runtime suites (60 tests).

*(PR by an AI agent, Claude Opus 4.5 via Claude Code.)*

> [!NOTE]
> ### Add stale-binding note to failed cell errors naming variables not updated before the failure
> - Adds `introspect.unreached_bindings` in [introspect.py](https://github.com/indexable-inc/index/pull/2531/files#diff-006579eb39a5fc5841047297950ecbb104d7094d82bdaec8fe1b9a4eeeda9cad), which statically analyzes cell source via `ast.parse` to determine which variable names were provably not assigned before the failing line.
> - Adds `runtime._stale_binding_note` in [runtime.py](https://github.com/indexable-inc/index/pull/2531/files#diff-a321ac036dbe5148143f01e84f705e2c20bda0e38f193e74b06c66889de27b95), which calls `unreached_bindings` and formats a `NOTE:` string listing up to 12 names (with `+N more` elision) appended to `job.error` on both exception and `KeyboardInterrupt` failures.
> - Adds unit and end-to-end tests in [test_stale_binding_note.py](https://github.com/indexable-inc/index/pull/2531/files#diff-2a12f4d7ad49007ca0618566d5503d5344512b3440f3b019f00f8260050274e1) covering AST edge cases and the elision behavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 32402c4.</sup>
> 
>

### DIFF
--- a/packages/mcp/default.nix
+++ b/packages/mcp/default.nix
@@ -3051,7 +3051,7 @@
         pkgs.fd
       ];
       strictDeps = true;
-      meta.description = "per-cell type check (ty) + issue #1754 bug 1-3 regressions + sh exit surfacing (#1766) + Result.value reachability (#2068) + find glob= filter (#1366) + in-band build stamp (#2110) + session-scoped job cancellation (#2104) + client-cancel interrupts in-flight run (#2387) + jobs.spawn ad-hoc awaitables (#2164) + grep files_only (#2246) + claude-history session search (#2245) + per-serve kernel trace file (#2355) + builtin shadow restore (#2430)";
+      meta.description = "per-cell type check (ty) + issue #1754 bug 1-3 regressions + sh exit surfacing (#1766) + Result.value reachability (#2068) + find glob= filter (#1366) + in-band build stamp (#2110) + session-scoped job cancellation (#2104) + client-cancel interrupts in-flight run (#2387) + jobs.spawn ad-hoc awaitables (#2164) + grep files_only (#2246) + claude-history session search (#2245) + per-serve kernel trace file (#2355) + builtin shadow restore (#2430) + stale-binding note on failed cells (#2526)";
     }
     ''
       export HOME=$TMPDIR/home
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Superseded by #2528, which landed the same #2526 fix (unexecuted_bindings + traceback note, tests in typecheckSmoke) while this PR's CI was blocked on the fleet outage. Only deltas here were the opposite tradeoff (skip enclosing-loop/try-unwind spans, under-warn) and a 12-name elision cap on the note; #2528 deliberately over-warns, so not re-litigating. Closing.

PR #2639, opened 2026-07-10
darwin: move computed paths into config

### LINKED ISSUE #2638: darwin: move computed internal paths out of option defaults
## Problem

`origin/main` at `d96784f6` fails `nix run .#lint` with two `no-mkoption-conditional-default` errors:

- `modules/darwin/mutable-files.nix:83`
- `modules/darwin/nfs.nix:57`

Both internal options compute a conditional `mkOption.default` from the submodule name. The lint requires the option declaration to stay self-contained and the computed value to be assigned from `config`.

## Reproduction

```sh
nix run .#lint
```

## Desired outcome

Move both computed internal defaults into module config without changing rendered paths.

### LINKED ISSUE #2640: index-delta: name indexed iterator item type
## Problem

The flake check on current `main` fails `index-delta-clippy`:

```text
packages/index-delta/src/diff.rs:254:77
error: return type contains an anonymous tuple
fn indexed(items: &[Value], from: usize) -> impl DoubleEndedIterator<Item = (String, &Value)>
```

Clippy runs with `-D clippy::anonymous-tuple-return-type`, so the iterator item needs a named type.

## Reproduction

Build the repository flake checks or the `index-delta-clippy` check.

## Desired outcome

Introduce the smallest named item type and preserve the diff behavior.

andrewgazelka:
## Summary

- move computed `_target` and `_mountpoint` values from conditional option defaults into submodule config
- remove the now-exposed unused `pkgs` argument from the NFS module
- replace the anonymous `index-delta` leaf tuple with a named type
- restore repository lint and `index-delta-clippy`

## Validation

- `nix run .#lint`
- `cargo test -p index-delta`

The host Cargo Clippy is older than the repository lint set, so the authoritative `index-delta-clippy` check is left to the Nix flake check and CI.

Closes #2638
Closes #2640

> [!NOTE]
> ### Move computed path resolution for darwin mutable files and NFS automounts into config assignments
> Moves the logic for computing `_target` (in [mutable-files.nix](https://github.com/indexable-inc/index/pull/2639/files#diff-8138811c3250ba97ceb05971140bf694aad1d713072d4193fe7e5e6887879268)) and `_mountpoint` (in [nfs.nix](https://github.com/indexable-inc/index/pull/2639/files#diff-2bb6df435ec5a186ad75732a7ba272d8c3c8cf2b060cf6396aa365043a6688ec)) from option `default` expressions into explicit `config` assignments. The resolved absolute path is now set unconditionally via config rather than as a default, which raises its merge priority. Also removes the unused `pkgs` parameter from the NFS module.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3995d78.</sup>
> 
>

### DIFF
--- a/modules/darwin/mutable-files.nix
+++ b/modules/darwin/mutable-files.nix
@@ -83,13 +83,14 @@
       _target = mkOption {
         type = types.str;
         internal = true;
-        default =
-          if lib.hasPrefix "/" name
-          then name
-          else "/${name}";
         description = "Resolved absolute target path.";
       };
     };
+
+    config._target =
+      if lib.hasPrefix "/" name
+      then name
+      else "/${name}";
   });
 
   files = lib.attrValues cfg.files;
--- a/modules/darwin/nfs.nix
+++ b/modules/darwin/nfs.nix
@@ -10,7 +10,6 @@
 {
   config,
   lib,
-  pkgs,
   ...
 }: let
   inherit
@@ -57,13 +56,14 @@
       _mountpoint = mkOption {
         type = types.str;
         internal = true;
-        default =
-          if lib.hasPrefix "/" name
-          then name
-          else "/${name}";
         description = "Absolute local mount point (the attribute name).";
       };
     };
+
+    config._mountpoint =
+      if lib.hasPrefix "/" name
+      then name
+      else "/${name}";
   });
 
   mounts = lib.attrValues cfg.automounts;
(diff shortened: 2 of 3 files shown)


andrewgazelka:
Retriggering pull request workflows after the latest pushed fix. (sent by an AI agent via Codex)

PR #2663, opened 2026-07-10
provenance: fix module dedup, launchd plist paths, whence lookup jank

wyattgill9:
Closes ENG-6624

Review of the `whence` eval-provenance stack (lib/provenance.nix + the two
`provenance.nix` modules + `packages/nix/whence`, shipped in #2423/#2426).
Ran an adversarial pass; found and fixed real jank in five places. All the
eval assertions and the four Nix lint stages (alejandra, astlog, deadnix,
statix) pass.

## Fixes

**Module dedup (correctness).** `modules/{home,darwin}/provenance.nix` are
functors `{provenance}: { ... }`. Applying the walker turns every import site
into a distinct *anonymous* attrset, which the module system cannot
deduplicate the way it dedups `imports = [ ./file.nix ]` by path. Combine a
bundled profile that pulls the module with `homeModules.provenance` (exactly
what a consumer does) and both instances declare `provenance.*`, so eval dies
with `The option 'provenance.enable' ... has already been declared`. Fix: an
explicit `key` restores dedup and `_file` restores error attribution; the
flake also shares a single `provenanceHomeModule` instance across its own
wiring sites. Regression test added (`tests/provenance.nix` now imports the
module twice).

**home-manager launchd plist path (correctness).** The home collector keyed
launchd agents as `Library/LaunchAgents/<attr-name>.plist`, but home-manager
deploys the plist named after the agent's `Label`
(default `org.nix-community.home.<name>`). The manifest key matched no
deployed file, so `whence ~/Library/LaunchAgents/...plist` on any agent fell
through to the deriver fallback. Now keys by `value.config.Label`.

**nix-darwin label guess (fail-loud).** The darwin collector hardcoded a
`org.nixos.<name>` fallback label. nix-darwin always sets
`serviceConfig.Label` (`mkDefault "${labelPrefix}.${name}"`), so the code now
reads it directly rather than silently diverging from a consumer's
`launchd.labelPrefix`.

**whence lookup (robustness).** `manifests` now also reads the pre-XDG
per-user profile manifest that older installs still use, and `main` trims a
trailing slash from `$HOME` so the `($home)/` prefix tests cannot miss every
home file.

**spurious settings chains.** A `programs.<x>.settings` chain whose only
definitions come from the declaring (wiring) module is the module populating
its own option, not the user configuring it. Such chains no longer claim a
settings site the user never wrote.

## Left alone (with reasons)

- **cwd-relative `whence` arg** - `path expand` resolving against cwd is
  standard shell semantics, not a bug.
- **`relativeTo` empty-base branch** - latent only; no current collector
  reaches it.
- **options.json cosmetic warning** - not worth a diff.

> [!NOTE]
> ### Fix module dedup, launchd plist paths, and manifest lookup in provenance
> - Adds explicit `key` and `_file` attributes to [modules/home/provenance.nix](https://github.com/indexable-inc/index/pull/2663/files#diff-40ac76caab7f7d53cb712be6b64a9d8068c15ab4bab93b52d250cb4d549febf8) and [modules/darwin/provenance.nix](https://github.com/indexable-inc/index/pull/2663/files#diff-179eea52238df67eb7e27dc0ba6fc908d263c1d4f8cea8f2f5f4523f88788c95) so the module system deduplicates instances and avoids "already declared" errors on repeated imports.
> - Fixes launchd plist manifest keys to use the configured `Label` (e.g. `value.config.Label` for Home Manager, `value.serviceConfig.Label` for nix-darwin) instead of the attribute name, so keys match deployed filenames.
> - Fixes `programs.*.settings` entries in `provenance.manifestFor` to only appear when the value is defined outside the declaring module, filtering out self-defined defaults.
> - Extends `whence` manifest discovery to include pre-XDG Home Manager profile paths and trims trailing slashes from `$HOME` during path comparisons.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8bc0856.</sup>
> 
>

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -437,6 +437,11 @@
       inherit indexPackages;
       portableServicesModule = ix.portableServices.homeModule;
     };
+    # One instance shared by every wiring site (the workstation profile and
+    # homeModules.provenance); the module's `key` also dedups the instances a
+    # consumer combines, but there is no reason to make them re-apply the
+    # walker.
+    provenanceHomeModule = import ./modules/home/provenance.nix {inherit (ix) provenance;};
     claudeCodeHomeModule = import ./packages/agent/home-manager/claude-code.nix {
       inherit indexPackages;
       promptModule = ./packages/agent/prompt;
@@ -458,7 +463,7 @@
       inherit indexPackages personalServicesModule ix;
       configRoot = personalConfigRoot;
       mutableFilesModule = mutableFilesHomeModule;
-      provenanceModule = import ./modules/home/provenance.nix {inherit (ix) provenance;};
+      provenanceModule = provenanceHomeModule;
       optionsModule = personalOptionsModule;
       indexSkillsSrc = paths.skills;
       tmuxModule = ./modules/home/tmux.nix;
@@ -577,7 +582,7 @@
       # file:line that defined them, and `whence <path>` reads it with zero
       # eval. Set `provenance.rev = self.rev or self.dirtyRev or null` in
       # the consuming flake. See modules/home/provenance.nix.
-      provenance = import ./modules/home/provenance.nix {inherit (ix) provenance;};
+      provenance = provenanceHomeModule;
       # Agent CLI modules: Home Manager is the user-facing configuration
       # surface, while the package wrappers remain the implementation detail.
       claude-code = claudeCodeHomeModule;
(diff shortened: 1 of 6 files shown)


andrewgazelka:
I am implementing #2672 from current origin/main and will add an independent `darwinModules.remote-builders` export in `flake.nix`. The diff should combine mechanically with this PR. I will keep the export isolated and rebase if this lands first. (sent by Codex)

Issue #1755, opened 2026-07-03
cross: Mac cannot substitute final CA cross outputs -- cache.ix.dev serves no /realisations (drv->output mapping)


andrewgazelka:
Follow-up to #1687 (whose eval-time fix works, see PR #1751): with the IFD closure now substitutable, a Mac gets through eval, but the FINAL cross output still does not substitute. At main 5ff28df8+, `nix build github:indexable-inc/index#dag-runner` on aarch64-darwin (with `--option narinfo-cache-negative-ttl 0`) substituted `cargo-units.nix` from cache.ix.dev, then planned a 96-drv LOCAL cross build and died on the x86_64-linux-only apple-sdk toolchain drvs.

Root cause, verified:

1. The workspace unit drvs (and the final `dag-runner-0.1.0` cross drv, `nv0gjd8zfxwpir04j7z2l8wyhzhjajnm-dag-runner-0.1.0.drv`) are floating content-addressed: `nix derivation show` gives `"outputs":{"out":{"hashAlgo":"sha256","method":"nar"}}` with a placeholder out path. Darwin eval cannot know the out path; it must resolve drv->output via the substituter's build-trace endpoint (`/realisations/<drv-output-id>.doi`).
2. cache.ix.dev serves no such endpoint: `/nix-cache-info` is 200, but `GET /realisations/<id>.doi` returns a bare `404 page not found` router miss. Its output NAR is pushed and 200s (`6gyjcj8c...-dag-runner-0.1.0` in the cache-push log) -- the bytes are there, only the drv->output mapping is missing.
3. Upstream: attic has no realisations/build-trace support at all, and NixOS/nix#6623 (closed not_planned) documents that `nix copy` of output paths does not propagate realisations either. Linux CI never notices: it can build, and `ix up` copies by path.

Candidate fixes (decide in this issue):
- Make the cross lane input-addressed (`contentAddressed = false` for the cross unit graph in `lib/rust/workspace.nix` `mkUnits`, cross-only): darwin eval then knows every out path and substitutes via plain narinfo. Bonus: cache-push's probe can finally skip already-pushed cross roots (today "CA roots have no eval-time path and always realise" every run). Cost: loses CA early-cutoff dedup for the cross lane only; native graph unchanged.
- Serve/push realisations: needs atticd+ncps endpoint work (nothing upstream to lean on) plus a registration-aware push (`nix copy` at installable level, not path level). Much larger blast radius.

[...]

PR #2680, opened 2026-07-10
cve-scan: differential export — scan only what changed (fixes never-completing runs)

wyattgill9:
## Problem

The CVE scan introduced in #2605 has never completed a live run. Both lanes recompute the world from scratch on every run:

- **Live lane**: one silent `nix build` of all 369 security roots. On the warm fleet runner a dry-run shows **4,569 derivations to build** — the CA rust workspace units can't be served by cache.ix.dev (#1755), so every run re-realizes them. The scheduled run sat 45 minutes with zero log output before being cancelled.
- **PR lane**: realizes the world **twice** (base + head) and re-evaluates **all of nixpkgs** per tree for the identity index. A dependabot PR run was still mid `nix-eval-jobs` when it was cancelled after ~14 minutes, leaving orphan eval processes on the runner.

## Fix: differential export keyed by drvPath

A root's eval-time `drvPath` is a precise fingerprint of everything that could change its output. The export now records `{root name → {drvPath, outPath}}` plus the path-info graph and final identity catalog as rolling state in `actions/cache`:

1. **One eval instantiates every root** (`nix eval --apply` over `securityRootPaths`, per-root `tryEval`): 72–106 s warm for all 369 roots (measured). No builds.
2. Roots whose drvPath matches the previous state and whose outPath is still in the previous graph are **reused without touching the store** — their graph closure is copied from the previous run's graph (closure-under-references is validated on both ends, so the copy is sound).
3. Only **changed roots** are realized, in one `nix build --keep-going` with **stderr streamed to the CI log** (kills the silent-45-minute failure mode). If the bulk build fails, each root is retried individually.
4. Identities are differential per store path: a path already in the previous catalog is reused verbatim (store paths are immutable); only new paths get `nix derivation show` + artifact inspection. This replaces the all-or-nothing identity cache key (`identity-cache-key.sh`, deleted).
5. The nixpkgs identity index is cached by locked nixpkgs revision and shared across both lanes.

**Add one package → the scan pays one eval + that package's closure.** Nothing else is recomputed.

## Failures become visible coverage debt, not sunk runs

A root that fails eval or realization no longer sinks the run: it lands in `unscanned-roots.json`, the run completes for everything else, and a non-empty unscanned list drives the existing `incomplete` terminal state (exit 3). The PR gate additionally blocks when head can't realize a root that base could — that's coverage regression the ratchet can't see.

## PR gate chaining

The gate restores the live lane's rolling state for the **base** scan, then feeds base's outputs as the **head** scan's previous state — a PR pays only for its own delta plus one nixpkgs eval on lock bumps (index cached by revision otherwise).

## Cache-key hygiene

The rolling-state key embeds `hashFiles('packages/cve-scan/src/**', 'Cargo.lock')`, so scanner changes start from scratch instead of reusing state written by different code. Missing/corrupt/evicted state degrades to a full export — it can never block a scan.

## Measured (wyatt-compute-1, warm store)

| Stage | Before | After (steady state) |
|---|---|---|
| Root realization | 4,569 drvs, silent, never finished | 0 builds when nothing changed (asserted via nix shim: zero `nix build`/`path-info` calls) |
| Root export (369-root eval + reuse) | n/a (never completed) | 107 s cold subset / 72 s differential |
| nixpkgs identity index | every run, both lanes, per tree | cached by nixpkgs revision, shared across lanes |
| Identities + artifacts | full re-inspection every run | 0 nix calls when nothing changed (early-return + skip) |

Steady-state live run ≈ eval (~2 min) + advisory fetch + scan ≈ **the 3-minute target**. Cold start (or scanner change) pays one full realization and self-heals into the differential regime.

## Known tradeoffs

- Scanner `src/**` or `Cargo.lock` changes reset the rolling cache → that run is a full re-realize/re-inspect (intended: state written by old code is not trusted).
- Reused identities keep evidence labels from the nixpkgs revision that first produced them; the store path contents are identical, so the advisory matching input is unchanged.
- Cold-start convergence assumes the persistent runner's warm `/nix/store`; a fresh runner's first run is expensive once.

## Validation

- 29 cargo tests pass, including 5 new differential tests (reuse-without-build, no-op run realizes nothing, broken root → unscanned, eval failure → unscanned, previously-inspected path skipped); clippy clean.
- x86_64-linux `ciChecks.rust-cve-scan.{alert-policy,ratchet-policy,cli-contract,clean-regression}` built from this branch on the fleet runner.
- Live-shape two-pass run on wyatt-compute-1: pass 2 with previous state made **zero** `nix build`/`nix path-info` calls (asserted by an injected `--nix` shim) and produced byte-identical roots/graph/state/identities.

No Terraform needed — this is a caching/CI-shape problem, not infrastructure provisioning.

> [!NOTE]
> ### Add differential scanning to cve-scan to fix never-completing runs
> - `export_roots` now reuses prior realizations and graph closures when `drvPath` is unchanged, only building changed roots, and records eval/build failures as unscanned roots instead of failing the entire export.
> - `export_identities` and `inspect_artifacts` accept a `--previous` catalog and skip re-processing store paths already present in it.
> - The CI workflow ([cve-scan.yml](.github/workflows/cve-scan.yml)) restores and saves per-run scan state caches (roots state, path-info graph, identities) keyed by nixpkgs revision and `Cargo.lock`; the PR gate ([pr-gate.sh](packages/cve-scan/ci/pr-gate.sh)) fails if the head introduces unscanned security roots not present in the base.
> - The `cache.rs` module and `identity-cache-key.sh` script are removed; identity caching is now handled via stateful differential reuse.
> - Risk: runs that cannot restore a previous state cache fall back to full re-export with a warning, which may still be slow on first run or cache miss.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ca6ad5f.</sup>
> 
>


andrewgazelka:
PR #2935 fixes the upstream Nix result-cardinality bug that currently makes the CVE gate report 378 results for 376 roots. Its gate must bootstrap with the branch-built patched Nix, which requires a small change in `packages/cve-scan/ci/pr-gate.sh`. I will keep that change limited to selecting `.#nix-ix` and passing `--nix`, so it should be straightforward to carry or drop when this differential scanner work lands. (sent by an AI agent via Codex)

andrewgazelka:
I am fixing #3008 in `packages/cve-scan/ci/pr-gate.sh`: preserve structured base failure evidence, continue through head evaluation, and keep head failures blocking. I will keep the change narrow; this PR may need a rebase after it lands. (sent by an AI agent via Codex)

PR #2876, opened 2026-07-11
fzf: yield Ctrl+R to atuin in base profile

TestingPlant:
fzf and atuin both bind Ctrl+R. Home Manager now warns on the double binding during evaluation of the root profile. atuin is the intended owner (see the fzf block comment), so drop fzf's Ctrl+R binding explicitly with historyWidget.command = "" rather than relying on source order.

> [!NOTE]
> ### Disable fzf's Ctrl+R history widget to yield the binding to atuin
> Sets `historyWidget.command` to an empty string in the base profile's fzf config, removing fzf's default Ctrl+R shell binding so atuin can handle history search instead. Affects Bash, Zsh, and Fish integrations in [base/default.nix](https://github.com/indexable-inc/index/pull/2876/files#diff-f41fc9d5d8fd7fc213d6f7fee2689b23cc1b16849c245dce044804a3431e2d8e).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0f89f86.</sup>
> 
>

### DIFF
--- a/modules/profiles/base/default.nix
+++ b/modules/profiles/base/default.nix
@@ -299,6 +299,12 @@ in {
           enableBashIntegration = true;
           enableZshIntegration = true;
           enableFishIntegration = true;
+          # Yield Ctrl+R to atuin (see above). fzf and atuin both bind
+          # Ctrl+R; atuin is sourced after fzf and wins, but Home Manager
+          # now warns on the double binding. Empty the history widget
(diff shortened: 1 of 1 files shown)


andrewgazelka:
Coordination: index#3069 will add `pkgs.codex` to the base profile package list only. I will avoid your touched sections and rebase if needed.

Issue #2587, opened 2026-07-10
mcp: recover nu engine after retained cwd is deleted


andrewgazelka:
## Problem

The persistent `nu()` engine retains a deleted worktree as its working directory. After `git worktree remove`, the next command fails before execution, even when the previous cell supplied a valid `cwd` for earlier commands.

## Reproduction

1. Run a `nu()` command with `cwd` set to a worktree.
2. Remove that worktree through `git worktree remove`.
3. Run another `nu()` command without immediately overriding the retained directory.

Observed:

```text
the engine's working directory `/Users/andrewgazelka/Projects/indexable-inc/index-issue-2583-mcp-format` no longer exists (a previous `cd` target was removed); pass cwd= to run somewhere real (it persists like `cd`), or nu.reset() for a fresh engine
```

## Desired outcome

When the retained working directory disappears, fail with a structured recovery signal or automatically use the explicit `cwd` supplied to the next call before validating the old directory. This would make worktree cleanup safe in persistent sessions.

(sent by Codex, an AI agent)

andrewgazelka:
Codex is driving this through a dedicated worktree. The exact reproduction confirms that an explicit `cwd=` already repairs the engine. The missing recovery is after a no-`cwd` call detects the deleted retained directory. I am implementing a typed stale-cwd exception and discarding only that failed engine, so the current call still fails loudly instead of redirecting while the following call starts fresh. This overlaps open PR #2573, so I am coordinating the base rather than racing its edits.

(sent by Codex, an AI coding agent)

PR #3143, opened 2026-07-13
fix(nu): probe SSH_AUTH_SOCK liveness per call, unset a dead socket loudly

### LINKED ISSUE #3140: kernel: stale inherited SSH_AUTH_SOCK breaks git commit signing mid-session
## Problem

The long-lived kernel inherits `SSH_AUTH_SOCK` from whenever it started. The 1Password agent socket path goes stale across agent restarts, so `git commit` in a kernel cell fails signing with `agent refused operation` / `failed to write commit object`, hours or days into a session, with no hint that the env is the culprit.

## Evidence

Session 2026-07-12: `git commit --amend` in a worktree failed signing until rerun with the known workaround prefix in the same `nu()` call:

```
hide-env -i SSH_AUTH_SOCK; git commit --amend -m '...'
```

(recipe lives only in auto-memory `kernel-stale-ssh-auth-sock-signing`, so every agent pays the failure once before recalling it)

## Proposed fix

Have the kernel (or the `nu()` bridge) resolve the signing agent socket per call instead of trusting the inherited env: drop/refresh `SSH_AUTH_SOCK` when the socket no longer accepts connections, or re-query launchd for the live 1Password agent socket. A cheap version: `nu()` probes `SSH_AUTH_SOCK` liveness once per call and unsets it when dead, which reproduces the manual workaround automatically and loudly logs the refresh.

Filed by an AI agent via session retro (Claude Code, Claude Opus 4.5).

andrewgazelka:
Fixes #3140.

**Problem:** the long-lived kernel inherits `SSH_AUTH_SOCK` once at startup while the agent behind it (1Password, launchd) restarts independently, so hours into a session `git commit` in a `nu()` cell fails signing with `agent refused operation` / `failed to write commit object`, with no hint that the env is the culprit. The manual workaround (`hide-env -i SSH_AUTH_SOCK; git commit ...`) lived only in auto-memory.

**Change** (`packages/nu-py/python/nu/__init__.py`, Python seam only, no Rust changes):

- `nu()` probes `SSH_AUTH_SOCK` liveness once per call: one `connect(2)` to the unix socket before the pipeline runs.
- When the socket is dead (ENOENT or ECONNREFUSED), the variable is unset in `os.environ` AND hidden in the engine's own env copy (the engine seeds its env from the host at construction). The hide is a compare-and-hide in one atomic eval, so a deliberate `env=` override never gets clobbered; the dead path crosses as a per-eval env var (no hand-rendered nushell quoting) and the sentinel hides itself in the same eval.
- The refresh logs loudly into the calling job's stdout: `[nu] SSH_AUTH_SOCK pointed at a dead agent socket (...)`.
- A live socket is never touched; once unset, later probes are no-ops.
- No re-pointing is attempted: with the variable unset, git's signing program (op-ssh-sign) resolves the live agent itself, and a replacement socket at the same path would have passed the probe. The pre-eval mirrors the caller's resolved `cwd`, so the deleted-PWD recovery contract (#2587) is preserved.

**Tests** (`packages/mcp/tests/test_nu.py`, runs in CI via `mcp.tests.nuTests`): dead socket file with no listener (ECONNREFUSED), missing path (ENOENT), live listener left untouched, and a deliberate `env=` override surviving host-sock death. The two dead-sock tests fail against unfixed `nu` (verified).

(sent by an AI agent via Claude Code, model Claude Opus 4.6)

> [!NOTE]
> ### Probe `SSH_AUTH_SOCK` liveness on each `nu` call and unset dead sockets with a log message
> - Adds `_dead_ssh_auth_sock` in [`nu/__init__.py`](https://github.com/indexable-inc/index/pull/3143/files#diff-1b1b715fd290d16b51363424dc714ebeb7b9a942d93a9bce548fb3d83d610fe4) that attempts a 1-second AF_UNIX connect to detect dead or missing socket paths.
> - Adds `_drop_dead_ssh_auth_sock`, called at the start of every `nu._run`, which removes a dead socket from both process env and the nushell engine env, then logs the path and issue number 3140.
> - Explicit `env` overrides passed via `nu.value` are preserved even when the host `SSH_AUTH_SOCK` is dead.
> - Adds four tests covering dead socket, missing path, live socket, and override-survives-dead-host-socket scenarios.
> - Behavioral Change: `SSH_AUTH_SOCK` is now silently removed from the engine and process env on every pipeline run when the socket is unreachable, rather than passing a stale path through.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 44aa5f6.</sup>
> 
>

### DIFF
--- a/packages/mcp/tests/test_nu.py
+++ b/packages/mcp/tests/test_nu.py
@@ -10,7 +10,11 @@ crossing, the df -> nu -> df roundtrip, the NuError diagnostic surface,
 import asyncio
 import datetime
 import inspect
+import os
 import pathlib
+import shutil
+import socket
+import tempfile
 
 import polars as pl
 import pytest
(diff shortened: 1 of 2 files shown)


andrewgazelka:
Closing as moot (Claude Code): the long-lived Python kernel this guards was retired in commit 24585304 ("Retire the Python ix-mcp server; mcp-ex is the only kernel"). packages/nu-py/python/nu/__init__.py now has no consumer on main (mcp-ex is Elixir), and the PR tests lived in the deleted packages/mcp/tests.

PR #3244, opened 2026-07-14
Nushell: link config from the mutable checkout

andrewgazelka:
## Summary

* discover every leaf under `users/andrewgazelka/config/nushell`
* deploy each leaf with `mkOutOfStoreSymlink` from the mutable index checkout
* preserve a writable Nushell data directory for history and plugin state
* isolate the derived file mapping in a focused Home Manager module

## Validation

* `nix run .#lint`
* built the overridden `andrewgazelka@hydra` Home Manager activation package
* realized the `config.nu` source and verified its terminal symlink target is `/Users/andrewgazelka/Projects/indexable-inc/index/users/andrewgazelka/config/nushell/config.nu`

Closes #3240

> [!NOTE]
> ### Link individual Nushell config files from mutable checkout instead of a single directory symlink
> - Replaces the recursive `Library/Application Support/nushell` directory symlink with per-file out-of-store symlinks, keeping the directory itself writable for Nushell runtime data.
> - Extracts Nushell file linking into a dedicated [nushell.nix](https://github.com/indexable-inc/index/pull/3244/files#diff-ef363f56adfb507f78ceeaac16eaa7129832f25a8584783918c10a0a3ab0af92) module that enumerates files under `${configRoot}/nushell` and creates a `home.file` entry for each.
> - An activation step removes any legacy generation-owned symlink at the Nushell data directory before `checkLinkTargets` runs, allowing the real directory to be created.
> - Behavioral Change: the Nushell config directory is now a real directory rather than a symlink, so Nushell can write runtime state (history, etc.) into it directly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 27dca3b.</sup>
> 
>


andrewgazelka:
Correctness review found that removed manifest entries currently leave their live targets behind, which would keep renamed or deleted Nushell code active. This PR will opt the discovered tree into `removeWhenUndeclared` after #3245 lands, then rebase and rebuild the activation package before merge.

PR #2678, opened 2026-07-10
darwin: add typed remote builders

andrewgazelka:
Fixes #2672
Refs #2670

Adds `darwinModules.remote-builders`, with one typed `nix.remoteBuilders` list driving both `nix.buildMachines` and an early system OpenSSH config snippet. The renderer fixes noninteractive transport, disables SSH multiplexing and agent identities, defaults to strict host-key verification, and points `NIX_SSHOPTS` at `/etc/ssh/ssh_config` so root's mutable SSH config cannot alter daemon builds.

Focused tests cover the empty default, exact build-machine translation, exact SSH rendering, host-specific ProxyCommand and keepalives, and duplicate-name rejection.

Validation:
- focused derivation build
- evaluation against the consumer's pinned nix-darwin `a1fa429`
- rendered config parsed by `ssh -G`
- Alejandra, Deadnix, Statix, and astlog on every changed file

The full repository lint still reports unrelated findings already present outside these files.

(sent by Codex)

> [!NOTE]
> ### Add typed remote builder configuration as a Darwin module
> - Adds [modules/darwin/remote-builders.nix](https://github.com/indexable-inc/index/pull/2678/files#diff-0bbebd27434d530cb3785197935d89f9c454f4b83c2a708482c1fd00484accac), a Darwin module that accepts `nix.remoteBuilders` as a typed list of builder records and derives coordinated `nix.buildMachines` and SSH config from them.
> - Each builder record includes fields for host, user, SSH key, supported systems, protocol (`ssh`/`ssh-ng`), and connection parameters; the module validates uniqueness of names, non-empty systems lists, and rejects store-backed SSH keys.
> - When `nix.remoteBuilders` is non-empty, the module enables distributed builds, sets `NIX_SSHOPTS` to use `/etc/ssh/ssh_config`, and writes an `ssh_config.d` fragment with Host stanzas that disable connection multiplexing.
> - Exports the module as `darwinModules.remote-builders` in [flake.nix](https://github.com/indexable-inc/index/pull/2678/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0) and adds tests in [tests/darwin-remote-builders.nix](https://github.com/indexable-inc/index/pull/2678/files#diff-ebb3d212ad42c1cc1af72710efc8bfeb16f21c46b94d7fc6ee3c3af8a8b73e28).
>
> 
>
[...]

PR #3329, opened 2026-07-15
darwin: gate protected Nix builders

andrewgazelka:
## Problem

An ordinary Darwin `nix build` can silently select a protected production worker through `/etc/nix/machines`. That happened repeatedly after protected builder use had been prohibited.

## Change

* Carries the reviewed typed remote builder module from #2678 onto current main.
* Adds an explicit `ambient` or `protected` access class. Only ambient records enter `nix.buildMachines`.
* Adds one Nix machine serializer in `lib/nix/build-machine.nix`.
* Installs `protected-nix-build` when protected records exist. It requires explicit builder names, a nonblank reason, and a bounded TTL.
* Runs one `nix build` with only the selected protected records and `max-jobs = 0`, so it cannot fall back locally.
* Audits authorization and terminal state, terminates the full client process group, then requires new daemon work to disappear from `nix store builds --json`.
* Preserves hardened SSH stanzas for both access classes.

## Validation

* Seven Python boundary regressions pass, including multi-builder fanout and timeout cancellation.
* `ruff`, strict `zuban`, and `pyright` pass.
* `git diff --check` passes.
* All changed Nix files parse with `nixfmt --ast`.
* No Nix command was run locally because this host's ambient builder configuration is the incident being fixed. CI owns the module evaluation and Alejandra gates.

## Coordination

* Supersedes stale draft #2678 without mutating its published branch.
* #3003 remains the source of the safe personal builder and is not duplicated here.
* The private Darwin consumer must classify `vc1-nix` as protected after this owner change lands.

Refs #3323

> [!NOTE]
> ### Add protected Nix builder module for Darwin with TTL-bounded remote builds
> - Adds a new Darwin NixOS module ([default.nix](https://github.com/indexable-inc/index/pull/3329/files#diff-2083be45a3934a38e9af407c2a5f0bb619a422837346ea770d2e26317de5cf9f)) that declares remote builders with two access modes: `ambient` (standard distributed builds) and `protected` (policy-enforced).
> - Ambient builders are wired into `nix.buildMachines` and `nix.distributedBuilds`; protected builders are excluded from distributed builds and require explicit invocation via a new `protected-nix-build` CLI.
> - The `protected-nix-build` tool ([protected_build.py](https://github.com/indexable-inc/index/pull/3329/files#diff-226dbd5e1783b476bfe6f5b6b46efdb018de68d06e0b42be096d966b6afe9c37)) enforces a JSON policy file, validates builder names, applies a TTL, handles SIGTERM/SIGKILL teardown, and emits structured audit records via `/usr/bin/logger`.
> - Per-builder SSH config stanzas are written to `ssh_config.d`, enforcing no multiplexing and strict identity use across all builders.
> - Unit and integration tests are added in [test_protected_build.py](https://github.com/indexable-inc/index/pull/3329/files#diff-cca8f0c38b53b9c3695392388080df657662f54aead32ca3a27763478cb1c74d) and [tests/darwin-remote-builders.nix](https://github.com/indexable-inc/index/pull/3329/files#diff-ebb3d212ad42c1cc1af72710efc8bfeb16f21c46b94d7fc6ee3c3af8a8b73e28) and registered in CI via [tests/default.nix](https://github.com/indexable-inc/index/pull/3329/files#diff-1cc580de297308d93d82f7b72446ae4b98832a8aae3378e9e134102519a0e33a).
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 36dc926.</sup>
> 
>


andrewgazelka:
Triage note (Claude Code, stale-PR sweep): still wanted, not moot. There is no `modules/darwin/remote-builders` and no `lib/nix/build-machine.nix` on main.

Onto current main the flake.nix half no longer applies: commit for #3899 moved the whole `darwinModules` composition out of `flake.nix` into `lib/home-modules.nix`, which now builds entries with `importApply (paths.modules + "/...")` and is re-exported as `inherit (homeSurface) darwinModules homeModules;`. The registration therefore has to move to `lib/home-modules.nix`, and the `tests/default.nix` entry has to be re-added around main new test list. The module itself is unlanded and has no equivalent anywhere on main, so it is not moot.

PR #3405, opened 2026-07-16
geyser: drop Determinate-only underscore literal

harivansh-afk:
## Problem

`modules/services/geyser/default.nix` set `default = 25_565;`. Underscore digit separators in numeric literals are a Determinate Nix / patched-nix extension, not standard Nix syntax. Stock Nix lexes `25_565` as `25` followed by the identifier `_565`, so any downstream flake that imports this module tree and forces the option fails:

```
error: undefined variable '_565'
       at modules/services/geyser/default.nix:159:21
```

CI passes because it evaluates with the patched Nix; consumers on stock Nix (e.g. `linux-ix` on hari-compute-1, which follows `index/nixpkgs` and forces the module set) cannot evaluate anything that forces this module.

## Fix

The reported literal is not the only one. This strips underscore separators from every genuine numeric literal in the **consumer-facing `modules/` tree** - the portability boundary a downstream stock-Nix evaluator actually parses - covering option defaults, forced config values, and examples across 11 files (geyser, velocity, minestom, minecraft + bedrock + voice-chat, spark, observability, inference, ray, darwin/ray).

Deliberate uses of the digit-separator dialect are intentionally left in place, because the repo implements this dialect on purpose (`packages/nix/nix/patches/0014-...accept-underscore-digit-separators...` plus the `rnix` patches under `lib/util/rnix-digit-separators/` that teach alejandra/statix/deadnix to format it). Untouched: the formatter fixtures under `packages/nix/`, the `lib/util/rnix-digit-separators` patches, `tests/`, and in-repo `examples/`/`users/` - all evaluated with the patched Nix, and the fixtures require the separators to test the formatters.

String/comment occurrences of the CUDA package-set identifier `cudaPackages_13_1` are not numeric literals and were left alone.

## Verification

- Reproduced the pre-fix failure: `nix-instantiate --parse modules/services/geyser/default.nix` fails on stock Nix 2.34.8.
- Post-fix: all 50 files under `modules/` parse cleanly with stock Nix 2.34.8 (also the repro version in the issue).

## Note for maintainers

This makes the `modules/` tree stock-Nix parseable, but the rest of the repo still uses the dialect by design. If stock-Nix portability of the whole tree is intended, a CI job evaluating the module set with stock Nix would keep it from regressing.

Fixes #3403

(sent by an AI agent via Claude Code, Opus 4.5)

> [!NOTE]
> ### Remove underscore digit separators from numeric literals across NixOS service modules
> Replaces underscore-separated integer literals (e.g. `25_565`, `19_132`) with plain integers across service modules including geyser, minecraft, velocity, spark, and others. Underscore literals are a Determinate Nix extension and are not valid in standard Nix, so this improves compatibility with upstream Nix evaluators.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 10ced92.</sup>
> 
>

### DIFF
--- a/modules/darwin/ray.nix
+++ b/modules/darwin/ray.nix
@@ -160,7 +160,7 @@
         objectStoreMemory = lib.mkOption {
           type = lib.types.nullOr lib.types.ints.positive;
           default = null;
-          example = 8_000_000_000;
+          example = 8000000000;
           description = ''
             Bytes of RAM for this node's object store (Plasma). Null lets Ray
             autodetect. The fabric env lifts Ray's 2GiB macOS cap
--- a/modules/services/geyser/default.nix
+++ b/modules/services/geyser/default.nix
@@ -100,7 +100,7 @@ in {
 
       port = mkOption {
         type = types.port;
-        default = 19_132;
+        default = 19132;
         description = "UDP port Geyser binds for Bedrock clients.";
       };
 
@@ -156,7 +156,7 @@ in {
 
       port = mkOption {
         type = types.port;
-        default = 25_565;
+        default = 25565;
         description = "Java server port Geyser connects to when address is explicit.";
       };
 
--- a/modules/services/inference/default.nix
+++ b/modules/services/inference/default.nix
@@ -139,7 +139,7 @@ in {
 
     port = lib.mkOption {
       type = lib.types.port;
-      default = 18_080;
+      default = 18080;
       description = "Port the server listens on.";
     };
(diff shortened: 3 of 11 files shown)


harivansh-afk:
Closing: out of scope for the work that surfaced it. Issue #3403 documents the problem for the index maintainers to fix as they see fit. (sent by an AI agent via Claude Code, Opus 4.5)

PR #3421, opened 2026-07-16
mcp: raise FabricError when the weave server is unreachable

andrewgazelka:
When the weave server is unreachable, `fabric.claude.session` / `fabric.run` surfaced a raw `httpx.ConnectError` traceback from `weave.put_blob`. This PR translates connection failures (`httpx.ConnectError`, `httpx.ConnectTimeout`) at fabric's submit boundary into `FabricError` with the operational handles:

```
FabricError: fabric: weave server unreachable (All connection attempts failed); health check: `curl -s http://127.0.0.1:7677/api/info`; restart: `launchctl kickstart -k gui/501/org.nix-community.home.weave-serve`
```

The health-check URL is derived from the request that actually failed, so a `WEAVE_URL` override shows the real target. Recording still fails loud; there is deliberately NO skip/spool fallback (that design is #3419).

Also cuts the kernel store flusher's per-retry `weave write failed, retrying` stderr spam to one loud line per reachable->unreachable transition (same handles in the message) plus one line on recovery; the backoff retry behavior itself is unchanged.

## Tests
- `test_run_weave_unreachable_raises_fabric_error`, `test_session_weave_unreachable_raises_fabric_error` (server absent: transport-level connect refusal; asserts the SDK client never spawns)
- `test_unreachable_logs_once_per_transition` (3 consecutive write failures then recovery: exactly one loud line, one recovery line)
- All 36 tests in `src/fabric/test_fabric.py` + `tests/test_store_auth.py` pass locally; live repro against a closed port produces the message above.

Note: draft PRs #3360 and #2807 touch the same regions (fabric submit path, `_writer` retry print) and will need a trivial rebase after this lands.

Fixes #3416

> [!NOTE]
> ### Raise `FabricError` when the weave server is unreachable during journal operations
> - Adds a `_journal` async helper in [`fabric/__init__.py`](https://github.com/indexable-inc/index/pull/3421/files#diff-4296e34ab4fac6c0d5c1e02c185abf6314d10be93f0ab536a5b7f7d61f588bd6) that catches `httpx.ConnectError` and `httpx.ConnectTimeout` and re-raises them as `FabricError` with an actionable message (curl health-check URL and `launchctl` restart command).
> - Wraps `weave.put_blob` and `weave.assert_facts` calls in both [`fabric/__init__.py`](https://github.com/indexable-inc/index/pull/3421/files#diff-4296e34ab4fac6c0d5c1e02c185abf6314d10be93f0ab536a5b7f7d61f588bd6) and [`claude.py`](https://github.com/indexable-inc/index/pull/3421/files#diff-a3385e17c91b79847e45af2e742bb2999d0bdcfa37e20a74f40e55d9428219a0) with `_journal`, so connectivity failures surface as `FabricError` instead of raw httpx exceptions.
> - Updates [`store.py`](https://github.com/indexable-inc/index/pull/3421/files#diff-a4763dded9f44ec2a563e568a62d900b996c6a956b6de36462468f1d2b03b976) background writer to log one detailed failure message per outage and one recovery message on success; repeated failures during backoff no longer spam stderr.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 56e8ef2.</sup>
> 
>


andrewgazelka:
Heads-up for sequencing with #3437 (issue #3419, steps 3+4 of #3418).

#3437 replaces the store's `_writer` thread and the fabric spawn-path weave RPCs with a local fsync'd spool (`weave.spool.Spool`) plus a background flusher: intent is durable locally before any subprocess spawns, and there is no synchronous weave call left on the spawn path. That means two pieces of this PR target code paths #3437 deletes:

- the `_journal()` FabricError wrapper around `put_blob`/`assert_facts` on the spawn path (those calls no longer exist there), and
- the one-loud-line handling in the store `_writer` (the writer thread is gone; the spool owns the once-per-URL unreachable/reachable-again latch).

The FabricError boundary is still valuable on the READ paths (`query`/`watch`/`wait`), which #3437 does not touch.

Whichever PR lands second rebases; happy to do the rebase on my side if this lands first. If you'd rather rescope this PR to the read paths now, that avoids the conflict entirely.

_sent by an AI agent via Claude Code_

PR #3470, opened 2026-07-16
cargo-unit: allow source-owned unit catalogs

andrewgazelka:
A committed Cargo-unit catalog lets trusted CI select concrete Rust root derivations with import-from-derivation disabled. The generated catalog remains a normal derivation for a cacheable drift check, while the existing unitsNix handle stays compatible.

The required gate now proves one real fixture root is enumerable with IFD disabled and all builders disabled. The fixture covers build, test, and bench target sets from one catalog.

Validation:
- nix run .#lint (10/10)
- strict no-IFD/no-builder drvPath evaluation
- Nix parse, shell syntax, formatter, and diff checks

> [!NOTE]
> ### Allow source-owned unit catalogs in `cargoUnit.buildWorkspace`
> - Adds a `unitCatalog` argument to `cargoUnit.buildWorkspace` in [`lib/rust/cargo-unit.nix`](https://github.com/indexable-inc/index/pull/3470/files#diff-82054df1a7d1648a5be88265641a7ae5d3b9e4ec4cdea72c518359ea4ae3f804) that accepts a Nix path or null; when supplied, the workspace imports that path instead of the generated catalog at evaluation time.
> - Adds `cargoUnit.generateUnitCatalog` as a recovery function that bootstraps generation using an empty catalog, so the catalog can be regenerated even if the source-owned file is syntactically invalid.
> - Introduces `mkUpdateCargoUnitCatalog` and `update-cargo-unit-catalog` flake tools (in [`flake.nix`](https://github.com/indexable-inc/index/pull/3470/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0)) to atomically install the generated catalog into the checkout, plus tests that validate drift between the source-owned and generated catalogs.
> - Fixes a hash collision bug in [`packages/nix/nix-cargo-unit/src/render.rs`](https://github.com/indexable-inc/index/pull/3470/files#diff-91be6a053ea25b59c072a514f469215dbbfeb6abbba3ddb8d77ad99ffa244e7f) where build-script run dependencies were excluded from unit identity, allowing distinct derivations to share the same attribute name.
> - Replaces evaluator-time `builtins.path` vendor staging in [`packages/nix/nix-cargo-unit/templates/units.nix.in`](https://github.com/indexable-inc/index/pull/3470/files#diff-22a77df39c2977097bc5e357af36873052c0aad50ca302d915b95ee50fdecc94) with a `pkgs.runCommand` derivation that symlinks vendor includes at build time, avoiding IFD.
> - The `update-rust-nightly` workflow is split into credential-free `prepare-update` and `publish-update` jobs; the prepare job now also regenerates the cargo-unit catalog and uploads it as a scoped artifact.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ed62c52.</sup>
> 
>


andrewgazelka:
Triage note (Claude Code, stale-PR sweep): still wanted in principle, but it cannot be rebased mechanically.

Three conflicts in `lib/rust/cargo-unit.nix` against the reworked `plannerSource` and `unitGraphJson` templating on main, plus the committed 13,241-line `tests/fixtures/cargo-unit-hello/unit-catalog` fixture, which is a snapshot of generator output that `packages/nix-cargo-unit/src/render.rs` has since changed. That fixture has to be regenerated before the test means anything. Left open for that work.

PR #3454, opened 2026-07-16
lint fix: rust lane (cargo fmt with the repo-pinned nightly)

andrewgazelka:
Implements #3433: a `rust` lane in the #3431/#3432 fixer pipeline.

## What
- `lintFix.lanes.rust`: `cargo fmt --all` run with the exact toolchain `rust-toolchain.toml` pins, plus the `rustfmt` component its component list omits. The pin's single statement stays in `lib/rust/tooling.nix`, now exported as `ix.repoRustToolchainFor`.
- Scoped fileset: tracked `.rs` files + every `Cargo.toml` (cargo-fmt discovers targets via `cargo metadata --no-deps`, manifests only: no Cargo.lock, no dependency sources, no network) + `rustfmt.toml` as `maybeMissing` so adding one starts scoping the lane instead of being ignored.
- Acceptance check grows a one-crate rust fixture: pre-fix `cargo fmt --check` must fail, the lane fixes it, post-fix `cargo fmt --check` passes from the same pinned toolchain, and refixing is byte-identical (no rustfmt lint stage exists, so the check gates with cargo fmt directly).
- `clippy --fix` is deliberately NOT here (that is #3434); it slots in before `cargo fmt` in this lane, per the formatter-last rule.

## Tree-wide reformat riding along
No cargo fmt gate existed before, so the tree had drifted: the lane's first patch was 208 files / 420KB. This PR applies it (a literal `git apply` of the lane-built `.#lint-fix-patch`), so post-merge main is fmt-clean and the lane realises to an empty patch, the no-op CA property the pipeline is built on. The clone-detect global ratchet is re-baselined 0.44 -> 0.45 with a written reason in `clone.toml`: rustfmt expands one-liners to multi-line shape, so fragments under `min_lines = 8` now cross it and count; no duplicated logic is added.

## Verified
- `nix run .#lint` 10/10 green on the reformatted tree
- `nix build .#lint-fix-patch` on the branch tip: 0 bytes (empty patch on a clean tree)
- `checks.x86_64-linux.lint-fix` acceptance gate: building on vin-compute-1 (in flight, will confirm before merge)

Closes #3433

🤖 Written by Claude Code (Claude Opus 4.6) for andrew@ix.dev

> [!NOTE]
> ### Add `cargo fmt` lint-fix lane using the repo-pinned nightly Rust toolchain
[...]

PR #3474, opened 2026-07-16
ci: push a style: autofix commit to PR branches from the lint fixer lanes

andrewgazelka:
Implements #3435: on PR CI, when the lint fixer lanes (#3431, #3450, #3454) produce changes, push a single `style: autofix` commit to the PR branch instead of failing the run.

## What changed

**`lib/per-system.nix`**
- `lintFix.appliedTo`: composes the full tracked tree with every lane's `fixed` output overlaid (the tree the repo becomes once the autofix commit lands).
- `checks.lint` now judges the verdict stages (astlog, filenames, dirnames, clone) over that **fixed** tree, per the #3431 design: fixable drift no longer fails the gate (CI commits it instead); genuinely unresolvable findings still do. The fixable stages keep running over the lanes' own output, which doubles as a live fixer-idempotence gate.
- New `lint-autofix` tool (`ix.writeNushellApplication`): applies the lane patch with `git apply --index`, commits as `style: autofix`, and pushes. Race-safe (pre-push `ls-remote` head check against the SHA the run checked out, plain non-force push, post-failure re-check to distinguish branch-moved from real failure) and loop-guarded (head commit already `style: autofix` while the fixed tree still differs means a non-idempotent fixer; the run fails loudly rather than pushing again).
- New flake check `lint-autofix` covering the tool end-to-end against local fixture remotes: empty-patch no-op, real push (commit parentage, subject, content), stale-head race skip, and the loop guard.

**`.github/workflows/check.yml` + `.github/scripts/push-lint-autofix.sh`** (minimal glue)
- On `pull_request` runs from same-repo branches, mint a `contents: write` token from the ix-mirror-sync app (the default `GITHUB_TOKEN` would not retrigger workflows on the pushed commit) narrowed via `permission-contents`, build `.#lint-fix-patch`, and run the tool from a clean worktree at the PR head SHA. The token travels via `GIT_CONFIG_*` env (never argv or on-disk config), and the checkout-persisted `GITHUB_TOKEN` extraheader is reset for the push so the app token actually authenticates it.
- Fork PRs and repos without the app configured get a loud failure message pointing at `nix run .#lint -- --fix` instead of a silent skip.

The pushed commit creates a new head SHA that goes through the normal required checks; nothing bypasses branch protection.

## Validation
- flake check `lint-autofix` green on aarch64-darwin and x86_64-linux
- `checks.lint` (fixed-tree gate) green on x86_64-linux
- `nix run .#lint` green on the branch
- this PR's own CI run exercises the workflow steps' no-drift path

Depends on the cache-masked lint fix that landed as 15990f63 (#3473).

🤖 Authored by Claude Opus 4.5 via Claude Code.

> [!NOTE]
> ### Push a `style: autofix` commit to PR branches from CI lint fixer lanes
> - Adds [push-lint-autofix.sh](.github/scripts/push-lint-autofix.sh) to build the lint-fix patch via Nix, apply it in a scratch Git worktree at the PR head SHA, and push a single `style: autofix` commit.
> - Extends the `flake-check` CI job in [check.yml](.github/workflows/check.yml) with two steps: minting an app token via `actions/create-github-app-token` (only on same-repo PRs with `MIRROR_APP_CLIENT_ID` set), then running the push script.
> - Introduces a `lintAutofix` Nushell application in [lib/per-system.nix](lib/per-system.nix) with safeguards: exits on empty patch, prevents idempotence loops if the head commit is already `style: autofix`, race-checks remote head SHA before pushing, and handles branch movement as a skip.
> - Adds a `checks.lint-autofix` derivation that validates push, skip, and loop-guard behaviors against local bare Git remotes.
> - Risk: if the app token is not minted (missing `MIRROR_APP_CLIENT_ID` or non-same-repo PR), the script fails the job rather than skipping silently.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 92cc83a.</sup>
> 
>

### DIFF
--- a/.github/scripts/push-lint-autofix.sh
+++ b/.github/scripts/push-lint-autofix.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# The composed fix patch for the checked-out tree (the PR's synthetic merge).
+# The required gate already built the fixer lanes (the lint check judges the
+# fixed tree), so this build only assembles the diff on top of cached lanes;
+# an empty patch is the common no-drift exit.
+patch="$(nix build .#lint-fix-patch --no-link --print-out-paths \
+  --option extra-experimental-features ca-derivations)"
+if [[ ! -s "$patch" ]]; then
+  echo "lint autofix: no fixable drift"
+  exit 0
+fi
+
+# The lint gate passed on the FIXED tree, so without a pushable fix this
+# drift would merge unrepaired; fail rather than go silently green. Token
+# minting is skipped for fork PRs and when the MIRROR_APP_* credentials are
+# absent (see check.yml).
+if [[ -z "${AUTOFIX_PUSH_TOKEN:-}" ]]; then
+  echo "fixable lint drift, but no autofix push token is available;" \
+    "apply the fix locally with: nix run .#lint -- --fix" >&2
+  exit 1
+fi
+
+# The patch is computed against the synthetic merge tree but the commit must
+# sit on the PR head, so apply it in a scratch worktree of the head SHA (the
+# merge commit's second parent, always present in this full clone). main is
+# lint-clean by induction (this gate), so merge-tree drift is head drift and
+# the patch applies; `git apply` inside lint-autofix is all-or-nothing, so
+# when main touched the same hunks the run fails loudly and rebasing the PR
+# is the fix.
+worktree="${RUNNER_TEMP:?RUNNER_TEMP is required}/lint-autofix"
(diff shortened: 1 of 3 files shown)


andrewgazelka:
Triage note (Claude Code, stale-PR sweep): still wanted, nothing on main supersedes it. `lib/per-system.nix:896` refers to "the CI autofix commit" as a thing that does not exist yet, and there is no `.github/scripts/push-lint-autofix.sh` on main.

Onto current main it conflicts in `.github/workflows/check.yml` and `lib/per-system.nix`, both of which moved substantially since this branch. It also touches `.github/**`, which the arming policy holds for a human read. Left open for a port plus that read.

PR #3475, opened 2026-07-16
lint fix: per-crate clippy --fix lane (#3434)

andrewgazelka:
Implements #3434: per-crate `clippy --fix` as a fixer lane in the lint fix pipeline (#3431), extending the lane seam from #3450/#3454.

## Design

**Per-package fix derivation (renderer).** `nix-cargo-unit` renders one `clippyFixByPackage.<pkg>` derivation per workspace package. Per-unit clippy invokes `clippy-driver` directly (no cargo), so cargo's `--fix` cannot be reused; the fix derivation replicates cargo fix's algorithm instead: run the fork driver with `--error-format=json --cap-lints=warn` and the same manifest-derived lint flags and prebuilt dep rlibs as the check gate, apply MachineApplicable suggestions with `rustfix` to fixpoint (4 passes, cargo's bound), and emit the fixed crate source. All of a package's units run sequentially against one shared writable copy so multi-target crates compose; each unit gets its own scratch root inside the derivation (store-copied OUT_DIRs are read-only, so a second target's copy into a shared scratch dir fails), the aggregated buildPhase rides `__structuredAttrs` (the single-env-var exec cap is 128 KiB), and fixup is disabled (patchShebangs would rewrite `#!/usr/bin/env` scripts in the fixed source tree, breaking byte-identity). Each of the three came out of running the lane over the real workspace, not the fixture.

**Incrementality via CA identity.** The fix derivation is content-addressed and its store name is exactly the scoped source's `builtins.path` name. A clean crate's identity fix therefore realises to the *same store path* as its source: unchanged crate + unchanged deps = cache hit = identity patch, and every downstream consumer dedups onto the existing graph. No whole-workspace `cargo clippy --fix` anywhere.

**Cross-crate re-verification, scoped to the enforced gates.** The rendered `units.nix` accepts `packageSourceOverrides`; `cargo-unit.nix` imports the same rendered graph once more with every package's fixed tree overriding its scoped source, so the ordinary per-unit clippy gates re-run against the composed workspace with dependents' rlibs rebuilt from fixed sources. The verify gate surfaces as `clippyFixVerified` next to `clippy` in `selectRootWithTests`' policyChecks, so it flows into CI through the same `passthru.tests` seam as the enforced gates and its scope matches them **by construction**. That scoping is load-bearing: CI's clippy gates cover only registry packages with `passthruTests`, and a never-enforced crate can carry live findings on main. `unibind` does (two pedantic `needless_pass_by_value`, not machine-applicable); an all-packages aggregate failed on it, and the differential on vin-compute-1 showed the ordinary unibind gate on the unfixed tree resolves to the identical resolved derivation and fails the same way, i.e. pre-existing residual, nothing the fixers changed. Because identity fixes share the source's store path, the verify graph's resolved derivations coincide with the primary clippy gates for clean crates: zero extra builds beyond the fix derivations themselves.

**Lane wiring.** The lint-fix rust lane's *input* is overlaid with each package's changed files before cargo fmt runs (formatter last); `lanes.rust` itself stays a generic tree transformer so the acceptance fixture keeps running it on synthetic workspaces. The overlay copies only changed-vs-original files and fails loudly when a changed file falls outside the lane fileset. Gated to x86_64-linux, where the autofix patch is produced.

## Checks

- `checks.x86_64-linux.cargo-unit-clippy-fix`: fixture workspace with a planted MachineApplicable `needless_return`; asserts the dirty crate comes out clean, the clean crate's fix is byte-identical, the identity fix realises to its source's store path (pinning the CA dedup the design leans on), and the fixture's own `clippyFixVerifiedByPackage` gates pass on the composed fixed tree. Green on vin-compute-1 at 7523703 (both halves plus the store-path identity assertion).
- `checks.x86_64-linux.lint-fix`: existing acceptance fixture, still green (lane stays generic).
- `checks.<system>.<pkg>-clippyFixVerified`: the per-crate re-verification gates over the real workspace.

## End-to-end on the real workspace (vin-compute-1)

All at 7523703 on vin-compute-1 (x86_64-linux):

- `cargo-unit-clippy-fix`: pass.
- All 109 `rust-*-clippyFixVerified` gates enumerated from `checks.x86_64-linux`: pass.
- `lint-fix` acceptance: pass (lane stays generic).
- `lint-fix-patch`: builds warm in ~15s and emits a real 17.5 KiB patch touching 11 files (genuine MachineApplicable fixes live on main today: `needless_lifetimes` in the google py bindings, plus dashboard-core, nu-py, search-py, tui, and the unibind conformance crates), while the enforced gates stay green because those findings are warn-level or sit in never-gated crates.
- Differential for the scoping decision: `unibind`'s ordinary clippy gate on the unfixed tree resolves to the same resolved derivation the old all-packages aggregate failed on (`unibind-clippy-0.1.0.drv`, two pedantic `needless_pass_by_value`, not machine-applicable), confirming a pre-existing residual on a crate CI never gates.

Three hardening commits came out of running the lane over the real workspace rather than the fixture: `__structuredAttrs` (aggregated buildPhase blew the 128 KiB single-env-var exec cap on file-search), per-unit scratch roots (store-copied OUT_DIRs are read-only, dashboard-core's second target failed to copy), and `dontFixup` (patchShebangs rewrote `#!/usr/bin/env` in blast-radius' test script, breaking byte-identity).

Issue: #3434. Design: #3431. Builds on #3450, #3454.

🤖 Written by Claude Opus 4.5 via Claude Code.

> [!NOTE]
> ### Add per-crate clippy --fix lane to the Rust lint pipeline
> - Adds a `clippyFixed` helper in [per-system.nix](https://github.com/indexable-inc/index/pull/3475/files#diff-9879a1f03396eb758ff538e44a00fb409b99bc0bbfac4656755b05716d036683) that, on x86_64-linux, overlays per-package clippy `--fix` outputs onto the source tree before the rust lane runs; other platforms pass sources through unchanged.
> - Extends [cargo-unit.nix](https://github.com/indexable-inc/index/pull/3475/files#diff-82054df1a7d1648a5be88265641a7ae5d3b9e4ec4cdea72c518359ea4ae3f804) to expose `clippyFixByPackage` (fixed source trees) and `clippyFixVerifiedByPackage` (clippy re-verification gates over the fixed sources).
> - Adds a `Driver::ClippyFix` mode in [render.rs](https://github.com/indexable-inc/index/pull/3475/files#diff-91be6a053ea25b59c072a514f469215dbbfeb6abbba3ddb8d77ad99ffa244e7f) that emits JSON diagnostics with `--cap-lints=warn` and drives a compile-then-apply fixpoint loop per package, producing a content-addressed fixed source derivation.
> - Adds an `apply-suggestions` subcommand to [main.rs](https://github.com/indexable-inc/index/pull/3475/files#diff-50450bb2844bd5aadba4f54593d42a14331704e3f8ebc3bfe70408fbac02781e) backed by [apply_suggestions.rs](https://github.com/indexable-inc/index/pull/3475/files#diff-2b33a899a69d7ecc448472f4811f1f27c3c0d9a0bdb0b0efd85f89b095ca5a1e), which applies `MachineApplicable` rustfix suggestions from JSON diagnostics to disk in reverse order, skipping overlapping or out-of-scope edits.
> - Adds an acceptance test fixture in [tests/fixtures/cargo-unit-clippy-fix](https://github.com/indexable-inc/index/pull/3475/files#diff-f089874a7246b6fafe01b5cb184030a56747d435053b983e575b2b51d2482350) with a two-crate workspace where the `dirty` crate triggers `needless_return`, verifying the fix is applied and re-verification passes.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7523703.</sup>
> 
>


andrewgazelka:
Triage note (Claude Code, stale-PR sweep): still wanted, not moot. Main explicitly reserves the seam for it: `lib/per-system.nix:983` says "`clippy --fix` (#3434) slots in BEFORE" the existing fmt lane.

It does not rebase mechanically. Onto current main the conflicts are `lib/per-system.nix`, `lib/rust/cargo-unit.nix`, `packages/nix-cargo-unit/src/apply_suggestions.rs` and `packages/nix-cargo-unit/templates/units.nix.in`: the fixer-lane code around `lanes` and `lint-fix-patch` was reworked after this branch, and `packages/nix/nix-cargo-unit` moved to `packages/nix-cargo-unit` in commit afabbe9a ("flatten package namespaces"). Needs a real port against the current lane shape.

PR #3499, opened 2026-07-17
fabric: render the tmux pane as a live human-readable transcript

andrewgazelka:
Closes #3496.

The `ix-agents` tmux window (#3478) gave every `fabric.claude.session` agent a pane, but it showed the CLI's raw stream-json. This makes the pane actually watchable:

- `fabric.render.Transcript`: one stateful line renderer; assistant text, tool calls with compact clipped inputs, tool results labeled by their `tool_use_id` and truncated, thinking lines, and a final `✔ done · 45.1s · 2 turns · $1.05`-style summary. Non-JSON and unknown events pass through verbatim/dimmed, so a CLI format change degrades to the old raw view, never a blank pane. Control frames and `stream_event` partial deltas (protocol, not conversation) show nothing.
- `pane_main` mirrors raw bytes to the SDK FIFO byte-exact (the transport is untouched) and renders per line to the pane; a render failure prints loudly and falls back to the raw line rather than severing the session.
- The pane command now carries the shim's `PYTHONPATH`, so the pane's `import fabric` resolves the same module tree as the shim (found live-testing: the tmux server env made the pane import a different fabric than the SDK side).

Validated: unit tests plus a pane end-to-end test (raw mirror byte-exact, rendered pane), `.#mcp.tests.strictTypecheck` and `.#mcp.tests.fabricTests` green, and a live `claude.session` on this machine whose tmux pane rendered the whole turn correctly.

> [!NOTE]
> ### Render the tmux pane as a live human-readable transcript of stream-json events
> - Adds a `Transcript` class in [render.py](https://github.com/indexable-inc/index/pull/3499/files#diff-904a18faafd7b77aed7ae4c6788d59a343bcb5f4ae12f2d384d10bd4c1ee88d8) that parses stream-json SDK/CLI events and emits ANSI-colored, human-readable lines; ignores control/stream_event frames, clips large payloads, and summarizes tool calls by name.
> - Updates `pane_main` in [tmux.py](https://github.com/indexable-inc/index/pull/3499/files#diff-c82b6a261843e9b2a3a81a5a60fe6a245e3f1bdf206886ae77c300d9c673d8de) to feed stdout lines through `Transcript` for display while keeping the raw byte-exact mirror to the SDK FIFO unchanged.
> - Fixes pane process launch to inherit the shim's `PYTHONPATH` so import resolution works correctly in the pane subprocess.
> - Adds tests covering transcript rendering, result summaries, passthrough of non-JSON, payload clipping, and end-to-end pane mirroring.
> - Behavioral Change: the tmux pane now shows a formatted transcript instead of raw JSON; the SDK-facing FIFO is unaffected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3b3d750.</sup>
> 
>

### DIFF
--- a/packages/mcp/src/fabric/fabric/render.py
+++ b/packages/mcp/src/fabric/fabric/render.py
@@ -1,0 +1,193 @@
+"""Human-readable rendering of the agent CLI's stream-json (index#3496).
+
+The tmux pane a ``fabric.claude.session`` CLI runs in (:mod:`fabric.tmux`)
+used to show the CLI's raw stream-json: machine framing, not something a
+human can follow. This module is the pane's display side:
+``Transcript.feed`` takes one raw stdout line and returns the text to
+show, so the pane reads as a live transcript (assistant text, tool calls
+with compact inputs, truncated tool results, a final result summary)
+while the SDK keeps the raw bytes through its own FIFO.
+
+Rendering never swallows information silently: a line that is not JSON,
+or a JSON event this module does not know, is shown as-is (unknown JSON
+dimmed and clipped), so a CLI format change degrades to the old raw view
+rather than a blank pane. Only frames that are protocol traffic, not
+conversation (SDK<->CLI control frames, ``stream_event`` partial deltas
+that a complete message always follows), render as nothing.
+"""
+
+from __future__ import annotations
+
+import json
+
+__all__ = ["Transcript"]
+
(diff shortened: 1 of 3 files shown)


andrewgazelka:
Closing as moot (Claude Code): commit 24585304 "Retire the Python ix-mcp server; mcp-ex is the only kernel" deleted packages/mcp/src/fabric, so the tmux pane renderer this PR adds has nothing to attach to on main.

PR #3542, opened 2026-07-17
mcp-ex: binary cell output must ride the wire, not kill it (fixes #3538)

andrewgazelka:
Fixes #3538.

A cell printing raw binary (`IO.puts(File.read!(<compiled binary>))`) killed the MCP connection: the harness hung forever and the job's history row was orphaned at `:running` (`Jobs.cancel` then raised "no such job"). Reproduced end-to-end in tests before fixing; the pinned death path:

1. `:unicode.characters_to_binary/1` answers invalid UTF-8 with an `{:error, ...}` **tuple**, which `IOProxy.convert/2` passed to the job's output buffer as if it were output.
2. `byte_size/1` on that tuple crashed the Job GenServer mid-`finish` -- after the eval had already succeeded -- before `History.finished/3` and before subscribers were notified.
3. The awaiting request task then died `:noproc`, and Stdio's `DOWN` handler silently dropped the pending reply: no response, ever, on a connection that looked alive.

The fix, layer by layer:

- **`IxMcp.UTF8` (new)**: valid UTF-8 passes byte-identical (the #3523 wire transparency, extended to output); invalid bytes become visible `\xNN` escapes, invalid codepoints `\u{...}`. `IOProxy` sanitizes both `:unicode` and `:latin1` `put_chars`, so the sink invariant (valid UTF-8 binary, always) holds no matter what a cell prints. `Job.append/2` now guards `is_binary` so any regression fails the offending `put_chars`, not the job record.
- **Stdio reply-always**: the request id is captured before dispatch, a handler task that dies abnormally is answered with a JSON-RPC `-32603` error, and a response the JSON encoder rejects degrades to an error reply instead of crashing the transport (and its linked read loop -- the whole connection).
- **History finalization**: `Jobs.History` is now a GenServer that monitors each recording job and drives its row to a terminal `:failed` if the process dies without reporting. `Jobs.cancel/1` on a dead-but-recorded id reports the recorded state instead of raising (including the async-Registry `:noproc` race the sandboxed suite caught).
- **Output cap**: one exec reply carries at most 64KiB of output, truncated on a codepoint boundary with a marker naming the original byte count; full output stays pageable via `Jobs.tail/lines/grep`.

The nix smoke test now round-trips a binary-printing cell over real stdio and proves the connection survives to answer the next request.

> [!NOTE]
> ### Fix MCP stdio transport to survive binary and invalid UTF-8 cell output
> - Adds [`IxMcp.UTF8`](https://github.com/indexable-inc/index/pull/3542/files#diff-af0cd410ef5f7c9c8af49a7aeba2052bf322e92eefd4db556e69685417251b14) with `sanitize/1` (escapes invalid bytes as `\xNN`) and `truncate/2` (byte-safe UTF-8 truncation); [`IOProxy.convert/2`](https://github.com/indexable-inc/index/pull/3542/files#diff-5aa2693b587d4194a0408dfff4e54d189e2f5656d509149e7b7246537ab1d670) now uses `sanitize/1` so all cell output is guaranteed valid UTF-8.
> - [`IxMcp.MCP.Stdio`](https://github.com/indexable-inc/index/pull/3542/files#diff-775ac658aaedab87f4cd5e80d254eaeb13c77ac446f0e4fb2583980ab94a7cb7) now tracks `ref -> id` for in-flight tasks; a crashed handler emits a JSON-RPC error reply instead of silently dropping the request, and a response that fails JSON encoding degrades to an error reply rather than killing the transport.
> - [`IxMcp.Jobs.History`](https://github.com/indexable-inc/index/pull/3542/files#diff-3ec5af6bd6a30c316cdf60f2fcbd7a3ef36b358e7e3682a10747c8035f6735e8) is refactored from Agent to GenServer and monitors job processes, automatically marking jobs `:failed` with elapsed time if they die without reporting.
> - [`IxMcp.MCP.Tools`](https://github.com/indexable-inc/index/pull/3542/files#diff-c1967b7edcd9f2a8628cd2ebf989cfe1e0daf0af2416367f04b4e816c897e776) truncates exec output exceeding `@max_output_bytes` using `UTF8.truncate/2` and appends an advisory line with byte counts.
> - Behavioral Change: `cancel/1` in [`IxMcp.Jobs`](https://github.com/indexable-inc/index/pull/3542/files#diff-4b32b41230aeb491b1cb281b931eee2f7be20df1fd5c07374727ce4819007dc7) now returns `{:error, status}` for gone-but-known jobs and raises only for unknown ids.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b45a9bd.</sup>
> 
>

### DIFF
--- a/packages/mcp-ex/default.nix
+++ b/packages/mcp-ex/default.nix
@@ -137,11 +137,13 @@
     }
     ''
       set +e
-      printf '%s\n%s\n%s\n%s\n' \
+      printf '%s\n%s\n%s\n%s\n%s\n%s\n' \
         '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18"}}' \
         '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}' \
         '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"session_set_name","arguments":{"name":"smoke"}}}' \
         '{"jsonrpc":"2.0","id":4,"method":"tools/call","params":{"name":"exec","arguments":{"intent":"utf8 wire roundtrip","budget":60,"code":"\"snow ☃\""}}}' \
+        '{"jsonrpc":"2.0","id":5,"method":"tools/call","params":{"name":"exec","arguments":{"intent":"binary output rides as escapes","budget":60,"code":"IO.puts(<<255, 97>> <> \"bin-marker\")"}}}' \
+        '{"jsonrpc":"2.0","id":6,"method":"tools/call","params":{"name":"exec","arguments":{"intent":"connection survives binary output","budget":60,"code":"\"alive-after-binary\""}}}' \
         | IX_MCP_ACTIONS_DB="$PWD/actions.db" ix-mcp-ex > response.jsonl 2> server-stderr.log
       rc=$?
       set -e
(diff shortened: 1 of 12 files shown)


andrewgazelka:
Merging with --admin per the infra-stall policy (2026-07-17): flake-check failed only on runner-queue admission ("worker waited 2895s for a runner slot, over the 1800s queue admission budget ... rerun once the backlog drains", indexable-inc/ix#7625), and closure-gate is its mirror. No code lane failed. The package's own CI derivations were built green locally before merge:

    nix build .#mcp-ex.tests.elixir   # compile --warnings-as-errors, format, credo --strict, ExUnit: 58 tests, 0 failures
    nix build .#mcp-ex.tests.smoke    # real stdio exchange incl. the new binary-output round-trip

PR #3543, opened 2026-07-17
mcp-ex: version the action log via PRAGMA user_version; route crash dumps into the state dir

andrewgazelka:
Fixes #3539.

## What happened

The 2026-07-17 22:46:25Z crash had the causality inverted in the issue body: an OLD #3513-era binary (flat `actions` schema with a `session` TEXT column) started against the NEWER normalized database that #3533 had already created on disk, failed to prepare its INSERT, and the failed `IxMcp.ActionLog` child took the whole application down. Its 3.6MB `erl_crash.dump` landed in the inherited cwd (`~/.config/nix`) and got committed by accident. Details: https://github.com/indexable-inc/index/issues/3539#issuecomment-5008501092

Old binaries cannot be fixed retroactively; this hardens current and future code.

## Changes

- **`PRAGMA user_version`-stamped ordered migrations** replace per-open column sniffing. The ladder: 1 = flat #3513 log, 2 = #3533 normalization, 3 = #3540 live rows. Unstamped files (always `user_version` 0 on disk today) are classified by inspection once -- fresh, flat v1, normalized v2, or current -- migrated stepwise (one stamping transaction per step, losslessly), and leave stamped.
- **A FUTURE `user_version` no longer kills the server.** A newer stamp means a newer server already ran against the file; the log opens disabled with a precise `Logger.error` naming found vs supported versions, writes are absorbed, reads answer empty, and the newer file is left untouched. Taking down every tool call over a log nothing on the hot path reads is the wrong blast radius.
- **`ERL_CRASH_DUMP` is exported into the state dir** (the action log's directory) before the supervision tree starts, so BEAM crash dumps stop landing in whatever cwd the MCP client ran the server from. The runtime reads the variable at dump time (verified empirically), so the app-start setting covers the incident class. The release smoke test now probes the routing end to end.

## Tests

- v1-flat fixture migrates losslessly and leaves the stamp (extended existing test)
- pre-live v2 fixture gains live columns and the stamp
- fresh database is created stamped; an unstamped current-schema file is stamped in place without a rewrite
- higher `user_version` disables logging with the precise error, keeps serving, leaves the file byte-identical

Validated locally exactly as CI runs it: `.#mcp-ex.passthru.tests.elixir` (compile --warnings-as-errors, format, credo --strict, 45 tests) and `.#mcp-ex.passthru.tests.smoke` both build green.

> [!NOTE]
> ### Version the action log schema via `PRAGMA user_version` and route crash dumps to the state directory
> - Introduces schema versioning in [`action_log.ex`](https://github.com/indexable-inc/index/pull/3543/files#diff-3e72adde29b4a1c32aaa66b83a963485a2f67735bd45ed8c1652532de4c4adab) using `PRAGMA user_version`, with ordered transactional migrations from v1→v2→v3 (v3 adds `status`, `stack`, and `stack_at` columns).
> - On startup, if the database was stamped by a newer server version, logging is disabled gracefully (GenServer state becomes `:disabled`) rather than crashing; API calls return neutral values.
> - Exposes `IxMcp.ActionLog.db_path/0` as a public function so other modules can resolve the action log path.
> - Routes Erlang crash dumps to `<action-log-dir>/erl_crash.dump` at application start via a new `route_crash_dumps/0` in [`application.ex`](https://github.com/indexable-inc/index/pull/3543/files#diff-bcb4fab743d7788d22bd4ea40d7b9569e8c1df5be6dd7207c7fd6a9b8c3e5ec7), unless `ERL_CRASH_DUMP` is already set or the DB is in-memory.
> - Behavioral Change: databases stamped with a future `user_version` silently disable action logging for that server instance instead of failing startup.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6809202.</sup>
> 
>

### DIFF
--- a/packages/mcp-ex/default.nix
+++ b/packages/mcp-ex/default.nix
@@ -137,13 +137,14 @@
     }
     ''
       set +e
-      printf '%s\n%s\n%s\n%s\n%s\n%s\n' \
+      printf '%s\n%s\n%s\n%s\n%s\n%s\n%s\n' \
         '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18"}}' \
         '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}' \
         '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"session_set_name","arguments":{"name":"smoke"}}}' \
         '{"jsonrpc":"2.0","id":4,"method":"tools/call","params":{"name":"exec","arguments":{"intent":"utf8 wire roundtrip","budget":60,"code":"\"snow ☃\""}}}' \
         '{"jsonrpc":"2.0","id":5,"method":"tools/call","params":{"name":"exec","arguments":{"intent":"binary output rides as escapes","budget":60,"code":"IO.puts(<<255, 97>> <> \"bin-marker\")"}}}' \
         '{"jsonrpc":"2.0","id":6,"method":"tools/call","params":{"name":"exec","arguments":{"intent":"connection survives binary output","budget":60,"code":"\"alive-after-binary\""}}}' \
+        '{"jsonrpc":"2.0","id":7,"method":"tools/call","params":{"name":"exec","arguments":{"intent":"crash dump routing probe","budget":60,"code":"System.fetch_env!(\"ERL_CRASH_DUMP\")"}}}' \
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Orchestrator note: #3542 merged first and both PRs touch packages/mcp-ex, so this went CONFLICTING (dirty PRs also skip pull_request CI, so the branch agent's check-watcher will never fire). Taking over the rebase onto main now; will re-push and merge on green.

PR #3545, opened 2026-07-18
agent: fail eval when systemPrompt.omitRules rides an explicit package

andrewgazelka:
Fixes #3537.

`programs.claude-code` / `programs.codex` fold `systemPrompt.omitRules` into `basePackage.override packageOverrides` only for their `mkDefault`-ed package. When a profile sets `package =` explicitly, that defaulted package is discarded and the option was silently ignored: the workstation generation built from 2c4d7d14 still shipped a `--system-prompt-file` containing the forceMerge and backgroundSubagents rule texts while `protectedMergeGuard = false` (set in the override args) already allowed force-merging -- a half-applied policy split across two surfaces.

Changes:
- **Both HM modules**: assert that `systemPrompt.omitRules != []` with an explicitly-set `package` fails eval loudly, using `options.programs.<name>.package.highestPrio` against the module's own `mkDefault` priority (no package instantiation just to compare).
- **workstation.nix**: thread `omitRules` into the `claudeCode` / `codex` override args (where the shipped packages are actually built) and drop the now-dead `systemPrompt.omitRules` module lines; prose comments updated to match the new wiring.
- **tests/default.nix**: claude-code counterpart of the existing codex omit-threading check, plus a guard-trip check (per-module `extendModules` fixpoint with an explicit package; home-manager checks assertions eagerly on config access, so `tryEval` failing is the guard tripping).

Validation:
- Workstation build (private HM config with `--override-input index` to this branch): `share/claude-code-launch-spec.json` -> `--system-prompt-file` no longer contains "Never bypass required checks" nor "harness subagent and task tools are absent"; codex `modelInstructionsFile` drops the forceMerge text while keeping backgroundSubagents (Codex genuinely has no harness subagent tools).
- Default `.#claude-code` build still contains both rule texts.
- Full private `homeConfigurations."andrewgazelka@hydra".activationPackage` evals with the branch.
- `checks.x86_64-linux.eval` (ix-eval-tests) evals green locally; `nix run .#lint` green; `nix fmt` clean on changed files.

> [!NOTE]
> ### Fail eval when `systemPrompt.omitRules` is set alongside an explicit package
> - Adds an assertion in the `claude-code` and `codex` Home Manager modules that triggers when `systemPrompt.omitRules` is non-empty but `package` has been set explicitly (i.e. at higher priority than `mkDefault`), emitting a clear error directing users to pass `omitRules` via the package override instead.
> - Updates [workstation.nix](https://github.com/indexable-inc/index/pull/3545/files#diff-92994d0c09d060a94c2474660b3f13f7ec27c73c00f02484fb737ac0eb467962) to move `omitRules` from the module option into the respective `claudeCode.override` and `codex.override` calls to comply with the new guard.
> - Adds tests in [tests/default.nix](https://github.com/indexable-inc/index/pull/3545/files#diff-1cc580de297308d93d82f7b72446ae4b98832a8aae3378e9e134102519a0e33a) that verify `omitRules` is threaded into the baked prompt when the default package is used, and that evaluation fails when an explicit package is combined with `omitRules`.
> - Behavioral Change: configurations that previously silently ignored `omitRules` when an explicit package was set will now fail evaluation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 31e2fc7.</sup>
> 
>

### DIFF
--- a/packages/agent/home-manager/claude-code.nix
+++ b/packages/agent/home-manager/claude-code.nix
@@ -13,6 +13,7 @@
 }: {
   config,
   lib,
+  options,
   pkgs,
   ...
 }: let
@@ -274,6 +275,21 @@ in {
       {
         assertion = cfg.systemPrompt.source == "house" || cfg.systemPrompt.omitRules == [];
         message = "programs.claude-code.systemPrompt.omitRules only applies when source = \"house\".";
+      }
+      {
+        # omitRules reaches the shipped wrapper only through defaultedPackage
+        # (basePackage.override packageOverrides); an explicit `package =`
+        # discards that override. Left unchecked this shipped a half-applied
+        # policy: the explicit package's permissions allowed force-merging
+        # while its baked prompt still forbade it (index#3537). The package
+        # stays defaulted only while no definition beats the module's own
+        # `lib.mkDefault defaultedPackage` (numerically lower highestPrio
+        # wins), so compare against that same mkDefault priority.
+        assertion =
+          cfg.systemPrompt.omitRules
+          == []
+          || options.programs.claude-code.package.highestPrio >= (lib.mkDefault null).priority;
+        message = "programs.claude-code.systemPrompt.omitRules is ignored when package is set explicitly; pass omitRules to that package's override instead (index#3537).";
       }
       {
         # The upstream module renders settings.json as a read-only store
--- a/packages/agent/home-manager/codex.nix
+++ b/packages/agent/home-manager/codex.nix
@@ -6,6 +6,7 @@
 }: {
   config,
   lib,
+  options,
   pkgs,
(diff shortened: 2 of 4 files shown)


andrewgazelka:
Merged with `--admin` under the explicit infra-stall authorization: flake-check sat queued for 25+ minutes on the saturated single-slot dispatcher (indexable-inc/ix#7625) with its 30-minute admission deadline already unreachable. The exact suite that gate runs (`checks.x86_64-linux.eval`, including the new dev-base assertions) passed locally at this head, as did `nix run .#lint`, the workstation + default `claude-code` build validations, and a full private HM `activationPackage` eval. (sent by an AI agent via Claude Code)

PR #3547, opened 2026-07-18
mcp-ex: sample the running cell's own line into the action log

andrewgazelka:
Fixes #3546.

While an exec runs, its job samples the eval process's stacktrace into the action log (#3540). The innermost formatted frame was nearly always stdlib machinery -- `(elixir 1.18.4) lib/process.ex:327: Process.sleep/1` for any sleeping cell -- and nothing recorded which cell line the run actually sat on.

Producer-side fix, all in mcp-ex:

- **Sample the cell's own line.** Workspace evals with `Code.env_for_eval(file: "cell")`, so frames from code the cell compiled (a module it defines) carry `file: ~c"cell", line: N` with N a 1-based line into the submitted source. Each sample now walks the raw frames for the deepest cell-owned frame and stores its line in a new `line` column (`update_stack/4`). Verified empirically: top-level statements and anonymous fns run interpreted through `:erl_eval` and leave no cell frame -- then `line` is NULL and a viewer falls back to the top frame.
- **Prune machinery from the stored stack.** The live sample now reuses `Evaluator.prune_stacktrace/1` (the Livebook-style cut already applied to error stacktraces), so the stored top frame is one the author can act on; an all-machinery stack keeps the raw frames rather than storing nothing.
- **Schema v4.** `PRAGMA user_version` 4 with a `3 -> 4` migration (`ALTER TABLE actions ADD COLUMN line INTEGER`), unstamped-file sniffing extended for the pre-line shape, `finish_action` clears `line` with the stack, `recent/2` entries expose it.

Note on the viewer half of #3546: nothing in this repo currently reads the `stack` column back (verified: `ActionLog.recent/2` is the only reader and is test-only), so there is no in-repo render to change. The `line` column is the contract a viewer needs, and the dashboard's `CodeBlock`/`Pane.line` already support live-line highlighting (the python MCP live-lines path) once an actions.db consumer exists.

Tested: `mix test` (63 tests -- new live cell-line test, pruning + fallback assertions on the sampling test, migration ladder/sniff/refusal tests updated for v4, new unstamped-v3-shape migration test), `mix format --check-formatted`, `mix credo --strict` against the shared config, `nix run .#lint`.

Generated with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.5)

> [!NOTE]
> ### Sample the running cell's source line into the action log during stack sampling
> - Adds a `line` column to the `actions` table in [action_log.ex](https://github.com/indexable-inc/index/pull/3547/files#diff-3e72adde29b4a1c32aaa66b83a963485a2f67735bd45ed8c1652532de4c4adab), populated during stack sampling with the deepest cell-owned frame's 1-based line number, cleared on finalize/cancel, and returned by `recent/2`.
> - Adds `ActionLog.update_stack/4` (previously arity 3), requiring callers to pass a line value (or `nil`).
> - Prunes plumbing frames from sampled stacktraces in [job.ex](https://github.com/indexable-inc/index/pull/3547/files#diff-cd26f23adf8d8e75fd1b4f9fd5f7cc906cd5aba85c413011e10200c5c85351c4) before storage, falling back to unpruned frames if pruning yields an empty list.
> - Makes `Evaluator.prune_stacktrace/1` public so it can be called from `Job`.
> - Bumps schema version to 4 with a v3→v4 migration that `ALTER TABLE`s the `actions` table to add `line`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d4ba066.</sup>
> 
>

### DIFF
--- a/packages/mcp-ex/lib/ix_mcp/action_log.ex
+++ b/packages/mcp-ex/lib/ix_mcp/action_log.ex
@@ -13,8 +13,10 @@ defmodule IxMcp.ActionLog do
   `cancelled` when its work completes (`finish_action/5`) -- for `exec` that
   is when the eval finishes, which for a backgrounded job is after the wire
   response shipped. While an exec runs, its job samples the eval process's
-  `current_stacktrace` into `stack`/`stack_at` (`update_stack/3`), so a
-  reader can show the line a hung cell sits on. There is deliberately no
+  `current_stacktrace` into `stack`/`stack_at` -- plus, when the stack has a
+  frame the cell itself owns, the 1-based cell source line into `line`
+  (`update_stack/4`, #3546) -- so a reader can show the line a hung cell
+  sits on and highlight it in the rendered cell source. There is deliberately no
   startup sweep of leftover `running` rows: several server instances share
   one database, so a sweep would clobber a live sibling's rows. A row whose
   kernel died stays `running` with a frozen `stack_at`; readers judge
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Merging with `--admin` under the standing infra-stall authorization (2026-07-17): the fleet CI dispatcher is not admitting Check jobs during the concurrency deploy (#3549 widened queue admission to 7200s for exactly this window; three branches' Check runs are queued with zero spawns for 50+ min, dispatcher journal on vin-compute-2 shows only status RPCs).

Evidence the change itself is green:
- First flake-check attempt: the `ix-mcp-ex` build **succeeded** (fixup completed) and the run then died to the 300s validation clock while `ix-cache-push-enqueue` reported "admission is full" -- every subsequent line in the runner-side log (`/var/log/ix-ci/runs/8970c10e.../Check/29623687116-1.stdout` on vin-compute-2) is `error: interrupted by the user`. A rerun has sat queued since 01:34:58Z.
- Local gates on the exact head: `mix test` 63/63, `mix format --check-formatted`, `mix credo --strict` (shared config), `nix run .#lint` all pass.

(comment by Claude Code, Claude Opus 4.5)

PR #3552, opened 2026-07-18
prompt: add moduleOptionShadowing rule from the index#3537 postmortem

andrewgazelka:
Closes #3551. Follow-up to #3537 / #3545.

#3537 was expensive to reverse-engineer because the failure mode is generic and silent: a module option folded only into a derived default (`package = lib.mkDefault (base.override packageOverrides)`) is discarded without error once the consumer sets the target explicitly. #3545 guarded the specific instance; this PR teaches the shared house prompt the pattern so the next agent recognizes it immediately.

Changes:
- **`packages/agent/prompt/rules.nix`**: one untagged rule, `moduleOptionShadowing` (the tag vocabulary is kind/runtime only, and nix-domain siblings like `gitBackedFlakeReferences` are untagged). Text covers both directions: author options against the final value or assert via `options.<ns>.package.highestPrio` compared with `(lib.mkDefault null).priority` (not `lib.modules.defaultOverridePriority`, the plain-definition priority 100), and suspect explicit-setting shadowing first when a module option seems ignored. Incident provenance cites index#3537 in `reason`, per house style.
- **system-prompt-eval**: one behaviors case (`module_option_shadowing` + task `ignored-module-option`): a hypothetical module option with clean eval and no effect on the built artifact, where the discriminating clue (an explicit `package` pin in the same config) is present but the mechanism must be inferred.

Validation: rendered `systemPrompt` via the import path in `packages/agent/prompt/default.nix` and reread the rule in context between `gitBackedFlakeReferences` and `fixAtSource`; both jsonl files parse and the task's `expects` id resolves; `nix run .#lint` green; `nix fmt` clean on the changed nix file.

> [!NOTE]
> ### Add `moduleOptionShadowing` rule to agent prompt rules
> Adds a new `moduleOptionShadowing` rule to [rules.nix](https://github.com/indexable-inc/index/pull/3552/files#diff-438311cd2350f776fff4331d71f04955c1d9d1d210df9c1a62e2980e444ffae5) based on the index#3537 postmortem. The rule explains that NixOS/Home Manager module options folded into a derived default can be silently discarded when an explicit package setting takes priority, and recommends applying options to the final value or using an eval-time assertion to catch the conflict.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e7c6c82.</sup>
> 
>


andrewgazelka:
Merging with `--admin` per the infra-stall allowance (Andrew, 2026-07-17): the one-slot CI dispatcher is saturated (ix#7625). Evidence: no Check run started between ~01:23Z and 01:58Z (the 00:58Z run alone held the slot ~60 min), and since 01:58Z the dispatcher has been idle 20+ min with three Check runs queued; this PR's `flake-check` has been QUEUED for over an hour without pickup. Nothing is failing, the required checks never ran.

Local validation standing in for the stalled checks:
- `nix run .#lint` green (alejandra, statix, deadnix, astlog, clone, ruff, filenames/dirnames)
- `nix fmt` clean on the changed nix file
- prompt renders green via both the import path and `nix eval --raw .#claude-code.systemPrompt` (rule asserts, duplicate-name and tag checks all exercised); new rule text present in the baked wrapper prompt
- `nix build .#system-prompt-eval.passthru.tests.{scoring,printsHelp}` green with the dataset additions; both jsonl files parse and the new task's `expects` id resolves

PR #3560, opened 2026-07-18
vmkit: raw block-device disks via VZDiskBlockDeviceStorageDeviceAttachment

andrewgazelka:
Closes #3554.

Every vmkit disk used to be an image file on APFS. This teaches the Virtualization.framework paths to attach a real host disk node straight to virtio-blk via `VZDiskBlockDeviceStorageDeviceAttachment` (macOS 14+), bypassing APFS sparse-file fragmentation, COW metadata, and double caching.

## Surface

- `--disk` on `boot-linux-gui` / `drive-linux` now repeats (first = boot disk, rest = extra virtio-blk devices, mirroring `boot-linux`) and each path may be an image file **or** a `/dev/[r]diskN[sM...]` node. Detection is by stat (block/char device), never path sniffing; a macOS bundle's `disk.img` symlinked at a device node works the same way in `boot-macos` / `run-macos` / `drive-macos`.
- `--disk-sync full|none` (default `full`) makes `VZDiskSynchronizationMode` explicit; `--force` overrides the mounted-filesystem refusal. Both only affect block devices.
- `--efi-vars` becomes required when the boot disk is a device (the `<disk>.efivars` default would land in `/dev`).
- `boot-linux` (libkrun) rejects device paths loudly: it opens disks by path, and no fd route exists on macOS (`F_SETFL` on a disk fd returns `ENOTTY`; reopening `/dev/fd/N` re-checks vnode permissions -- lima-vm/lima#5104).

## Safety

Before a read-write attach, the device is refused while it hosts mounted filesystems unless `--force`: a `diskutil list -plist` walk that also follows APFS physical-store linkage (an APFS container carved onto `disk0s2` mounts its volumes under a synthesized `disk3`, so a naive mount-table scan of `disk0` sees nothing while `/` lives on it).

## Privilege

`/dev/disk*` is root-owned and the VM must not run as root. When the direct open hits `EACCES`, vmkit re-runs itself under plain **interactive** sudo as a hidden `open-block-device` subcommand that opens the node and passes the fd back over a private unix socket (`SCM_RIGHTS`); the fd lands in an `NSFileHandle` for the attachment. No sudoers entry is ever installed -- lima's `NOPASSWD` helper entry was a root file-open oracle for every local user (lima-vm/lima#5112) and got their feature reverted (#5113). Hardening follows their #5117 redo: the helper only opens literal disk nodes (grammar-validated, `O_NOFOLLOW`, no symlinks), requires the sudo context (`SUDO_UID`) and a `0700` socket directory owned by the invoking user; the parent accepts only a root peer (`getpeereid`) and revalidates the received fd (`fstat` type + `st_rdev`) before VZ sees it.

## Validation (on hydra, Apple M5 Max, macOS 26.5)

- `nix run .#lint` clean; `cargo fmt --check` clean; 28 unit tests pass (`cargo test -p vmkit`), including an SCM_RIGHTS handoff roundtrip over a socketpair and the diskutil-plist mount matching (APFS linkage fixture from this machine).
- Functional roundtrip, direct-open path: `hdiutil attach -nomount ram://` ramdisk as second disk of a purpose-built NixOS guest (repart image, UKI); host writes a marker at sector 0, the guest copies it to sector 8 (guest read) and writes its own at sector 4 (guest write), both read back on the host. Green.
- Same roundtrip with the ramdisk node `chown root:wheel` (devfs): direct open denied, sudo helper handed the fd across, guest roundtrip green. The first run of this lane caught a real bug live: `tempfile` dirs default to 0755 and the **helper refused its own parent's socket dir** -- exactly the designed defense; fixed by creating the handoff dir 0700 explicitly.
- Mounted-filesystem refusal: mounted ramdisk (whole disk and slice) refused; internal `disk0` refused listing `/`, `/nix`, etc. through the APFS physical-store walk; after `diskutil unmountDisk` the attach proceeds; `--force` skips the check (the kernel's exclusive-open then surfaces `EBUSY` loudly if a mounted fs still holds the node).
- libkrun rejection and the `--efi-vars` requirement both error loudly before any device is touched.

## Not in this PR

- The A/B fio benchmark (raw partition vs preallocated image on APFS) on a real internal partition is a follow-up pending explicit human sign-off -- repartitioning hydra's internal SSD is destructive (issue #3554 notes this gate).
- The darwin `passthru.tests` / clippy gate is broken on main independently of this change (`cargo-unit-test-manifest` aborts listing `unibind`: dyld cannot load `@rpath/libstd-*.dylib`); verified identical on `main`. Clippy was run via the stock-toolchain preflight (fork lints stripped) with the vmkit-scoped output clean.

Written by an AI agent (Claude Code, model Fable 5), driven end to end including the live validation above.

> [!NOTE]
> ### Add raw block-device disk support via `VZDiskBlockDeviceStorageDeviceAttachment` in vmkit
> - Adds a new [blockdev.rs](https://github.com/indexable-inc/index/pull/3560/files#diff-f989b7927c824996b8df5342c7f514f63ec4eca54f8e301d011e725f3348860c) module that detects device nodes by `stat`, opens them with `O_NOFOLLOW`, and builds a `VZDiskBlockDeviceStorageDeviceAttachment` with a configurable sync policy (`--disk-sync`, `--force`).
> - Refuses to attach any device that has mounted filesystems (including APFS containers and snapshots detected via `diskutil -plist`) unless `--force` is passed.
> - Root-owned devices are opened via a hidden `open-block-device` subcommand invoked through `sudo`, which passes the fd over a private Unix socket using `SCM_RIGHTS`; the VM process itself is never elevated.
> - Linux VZ guests (`boot-linux-gui`, `drive-linux`) now accept multiple `--disk` arguments, each of which may be a raw device node or an image file.
> - macOS VZ guests (`run-macos`, `drive-macos`, `boot-macos-screenshot`) accept the same disk policy flags when `disk.img` in a bundle is a symlink to a device node.
> - libkrun-backed `boot-linux` explicitly rejects device node paths with a clear error.
> - When the boot disk is a device node, `--efi-vars` becomes required since there is no bundle directory to derive a default path from.
> - Risk: interactive `sudo` prompt appears mid-boot when opening a root-owned block device; cancelling leaves the VM unstarted.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 849cc2c.</sup>
> 
>


andrewgazelka:
Merging with `--admin` under the infra-stalled-checks allowance: flake-check's first attempt was killed by the 300 s validation clock while rebuilding the cargo unit-graph IFD (this PR touches workspace Rust sources), the `ci/big-change` label was added and all jobs rerun, and the rerun has now been queued for 65+ minutes while the shared runner serves newer runs (gate-primary retargeting was in flight for part of that window). Local gates are green: `nix run .#lint`, `cargo fmt --check`, `cargo check --all-targets`, 28 unit tests, a vmkit-scoped clippy preflight, and the end-to-end hardware validation described in the PR body (direct-open and sudo fd-handoff roundtrips, mounted-refusal matrix). (comment by an AI agent via Claude Code)

PR #3642, opened 2026-07-19
nix fork: fetch github inputs via git when submodules are requested

andrewgazelka:
Implements #3626: patch our vendored nix so `github:` (and `gitlab:`/`sourcehut:`) flake inputs work when submodules are requested. Adds patch 0023 to the nix series.

## Design

Candidate shapes considered were (a) a fetch-time branch inside the archive scheme, (b) redirection at input construction, and the combination the code actually demands. The chosen shape is construction-first:

- `GitArchiveInputScheme::inputFromAttrs()` constructs the equivalent `git` input (`type=git`, `url=https://<host>/<owner>/<repo>.git`, same ref/rev, `shallow=1` to match no-history tarball semantics) whenever `submodules=1` or `lfs=1` is requested, delegating through `Input::fromAttrs`. Construction is the correct layer because flake.lock then records a plain `git` node that stock nix consumes; a lock recording a `github` node with a `submodules` attribute is rejected by stock nix on the next read, which is exactly the NixOS/nix#13571 failure. A pure fetch-time branch (shape a) would have written such locks.
- One bypass exists: flake `self` attrs are injected into an already constructed input (`applySelfAttrs`), the NixOS/nix#14982 silent-empty-dirs path. `GitArchiveInputScheme::getAccessor()` performs the same redirect as a backstop, and `getFlake()` erases `__final` next to the `narHash` erase its own FIXME already calls for; otherwise `checkLocks()` pins the github `type` field-by-field on the refetch.
- `Input::fromURL/fromAttrs` keep a scheme set by a delegating scheme (2 lines), the archive URL parser accepts `submodules`/`lfs` query params it previously dropped silently, and explicit `false` is normalized away so no lock ever carries the attr on an archive node.

Precedent: in-tree `GitHubInputScheme::clone()` already maps github to `git+https`; roberth names this option in NixOS/nix#14982. His export-ignore/export-subst caveat (tarball vs git checkout content can differ) is documented in the patch, the attr docs, and the rl-next note; behavior only changes when submodules/LFS are requested, cases that were previously broken. Auth caveat likewise: private submodule fetches use git credentials, not forge `access-tokens`, as `git+https` inputs always have.

Upstream intent: `hold` (humans submit nix patches per NixOS/nix#15984); no upstream implementation existed to vendor as of 2026-07-19.

## Validation (before = current fleet nix 2.34.7+ix.p22, after = this branch's `nix-ix` 2.34.7+ix.p23)

Repro from NixOS/nix#14982, `nix flake prefetch 'github:jakubgs/nix-submodule-bug-repro?submodules=1'`:

- before: silently drops the param; tarball hash `sha256-zfCsTa...`, `dummy-submodule/` is an empty directory.
- after: fetches via git (`submodules: true`, `shallow: true`, `type: git` in the lock metadata), tree contains `dummy-submodule/README.md` ("Dummy file for Nix Flake submodule test.").

Attr-form input (`type = "github"; submodules = true`): before, `error: input attribute 'submodules' not supported by scheme 'github'`; after, locks cleanly to a git node.

Regression: plain `github:jakubgs/nix-submodule-bug-repro` prefetch is byte-identical before/after (`sha256-zfCsTa2yeEwCoBXSkJuSwZisOiEGP5W/3YzC1V9//ds=`, same store path).

Lock compatibility (the NixOS/nix#13571 transitive scenario, `github:pedorich-n/playit-nixos-module` which sets `inputs.self.submodules = true`): patched nix locks the dependency as `{type: git, url: ..., submodules: true}` while `original` stays the user's plain github ref. The pre-patch binary then re-runs `nix flake lock` as a no-op (the step that previously errored) and fetches the locked input with narHash verification green.

Unit tests: 9 new construction/redirect cases plus 2 URL-form tests inside the patch (`src/libfetchers-tests/input.cc`); full fetchers (29) and flake (21) suites pass. Fast gates `patched-src-nix` and `patch-dag-nix` green; full `.#nix-ix` builds on aarch64-darwin.

Closes #3626.

(PR authored by an AI agent via Claude Code, Claude Opus 4.5)

> [!NOTE]
> ### Fetch forge inputs via git when submodules or LFS are requested in Nix fork
> - Adds [patch 0023](https://github.com/indexable-inc/index/pull/3642/files#diff-2c062334644238a839dd47613c8445c2384e62c1a9f14c5c3b5d719cfb078568) to the forked Nix build, which rewrites forge archive inputs (GitHub/GitLab/Sourcehut) that request `submodules` or `lfs` as equivalent `git+https` inputs in `libfetchers`.
> - The redirect happens in `getAccessor` as a backstop and also clears `__final` on refetch to handle self attrs correctly.
> - Updates [fork-packages.nix](https://github.com/indexable-inc/index/pull/3642/files#diff-35e5ff5b02c2dcd32c93fd770ad2194f9fcbfc9f35fdcbd56821891b6e7452f4) and [dag.json](https://github.com/indexable-inc/index/pull/3642/files#diff-bb8cabe3c8aca5a21a04124e99377308810f2ffd5701104ed528dd747207dc2d) to apply patch 0023 after patch 0013.
> - Adds a [site update post](https://github.com/indexable-inc/index/pull/3642/files#diff-7e2ca912d964810cd256571e271e8875adc66c4f3612a6724f3a846988f3217b) documenting the change, lock-file effects, and caveats.
> - Behavioral Change: lock files for forge inputs using `submodules` or `lfs` will now record a `git+https` URL instead of an archive URL.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bb40fab.</sup>
> 
>

### DIFF
--- a/lib/fork-packages.nix
+++ b/lib/fork-packages.nix
@@ -543,6 +543,16 @@
           upstream = "hold";
           reason = "Fleet CI policy for indexable-inc/index#3317. Validate the process-aware deadline before proposing a general Nix interface; humans submit Nix patches upstream per NixOS/nix#15984.";
         };
+        # 0023: forge archive inputs (github:/gitlab:/sourcehut:) requesting
+        # submodules or LFS are constructed as the equivalent git+https input
+        # (archive tarballs cannot contain that data), so lock files record a
+        # plain `git` node stock Nix understands. Fixes the hard failure of
+        # NixOS/nix#13571 and the silent empty-submodule trees of
+        # NixOS/nix#14982; the mapping mirrors GitHubInputScheme::clone().
+        "0023-libfetchers-fetch-forge-inputs-via-git-when-submodul.patch" = {
+          upstream = "hold";
+          reason = "Implements roberth's implicit git+https switch from NixOS/nix#14982 (also fixes NixOS/nix#13571; indexable-inc/index#3626). Hold: humans submit Nix patches upstream per NixOS/nix#15984.";
+        };
       };
     }
   ];
--- a/packages/nix/nix/patches/0023-libfetchers-fetch-forge-inputs-via-git-when-submodul.patch
+++ b/packages/nix/nix/patches/0023-libfetchers-fetch-forge-inputs-via-git-when-submodul.patch
@@ -1,0 +1,544 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Andrew Gazelka <andrew@ix.dev>
+Date: Sun, 19 Jul 2026 00:43:38 -0700
+Subject: [PATCH] libfetchers: fetch forge inputs via git when submodules are
+ requested
+
+The github/gitlab/sourcehut input schemes download `git archive`
+tarballs, which never contain submodule content or Git LFS data, so
+`submodules = true` on such an input either failed at input
+construction with "input attribute 'submodules' not supported by
+scheme 'github'" (NixOS/nix#13571) or, when injected after
+construction by flake `self` attributes, was silently ignored and
+produced a tree with empty submodule directories (NixOS/nix#14982).
+roberth's triage of #14982 names the fix implemented here: implicitly
+switch to the git+https input scheme. The mapping already exists
+in-tree: GitHubInputScheme::clone() clones exactly that URL.
+
+Redirect at input construction, not at fetch time:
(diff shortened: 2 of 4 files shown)


andrewgazelka:
Superseded: the identical patch content landed via #3634 (f9df9760); restricted to the patch files this branch is byte-identical to main, and the visible PR diff was the three-dot merge-base view. (comment by an AI agent via Claude Code, Claude Opus 4.5)

PR #3754, opened 2026-07-19
nix fork: abort daemon workers with their client, prune zombie-owned status entries

andrewgazelka:
Fixes #3752 (dead-client daemon wedge; sibling of #3559's downloader half).

## Incident (hydra, 2026-07-19, daemon 2.34.7+ix.p29)

A bot-run nix client was SIGKILLed. Ten and a half hours later, `nix store builds --json` still reported **33 builds "in flight", owned by three daemon workers that existed only as `(nix)` zombies** (27 by pid 47606, 5 by 8745, 1 by 68961) under an unreapable parent; new builds of the same drvs (`strsim` et al.) sat "building" 55 minutes with flat logs; killing another client mid-incident minted a second batch of phantoms. Only `launchctl kickstart -k system/org.nixos.nix-daemon` cleared it.

## Root cause, in the fork source

1. **Lost interrupt wakeups** (patch 0031). Client death fires `MonitorFdHup` -> `triggerInterrupt()`, which sets a flag and forwards a one-shot SIGUSR1 (`ReceiveInterrupts`). `Worker::waitForInput` (src/libstore/build/worker.cc:399 pre-patch) polls only the children's log fds, with **no timeout** under our daemon config (`max-silent-time = 0`, `timeout = 0`, `min-free = 0` -- verified live on hydra). A trigger landing between the last `checkInterrupt()` and the `poll()` syscall is lost; with every builder silent that poll never returns, and the worker outlives its client indefinitely while holding goals, builders, build users, and path locks. One lost race = one permanent wedge; the same mechanism made #3559's children TERM-immune. Fix: a self-pipe written by an interrupt callback, registered in the poll set -- level-triggered, cannot be lost. (Checked upstream master for a cherry-pick first: its `Waker` self-pipe serves cross-thread goal wakeups only and is not wired to `triggerInterrupt`; NixOS/nix#15691 covers the Linux POLLERR spin + fork signal-thread restart, both already handled or inapplicable here.)

2. **Zombie-blind staleness** (patch 0032). `nix store builds` pruned entries whose writer failed `kill(pid, 0)` -- which succeeds for zombies, exactly what the incident's three workers were. Writers now hold a lifetime `flock` on their entry, acquired on the temp file before the rename so no reader can catch an unlocked live entry; the kernel releases it on **any** death (SIGKILL, crash, zombie), and pid reuse can't fake liveness. Legacy entries keep the pid probe, now zombie-aware on Linux (`/proc/<pid>/stat` state) and Darwin (`KERN_PROC` `p_stat`), so mixed-version fleets shed corpses too.

## Receipts

Repro on a throwaway daemon (private store/state/socket; script exercised both halves):

Before (p29): SIGKILL the worker with its parent stopped ->
```
-- worker after SIGKILL with stopped parent: Z    <defunct>
-- nix store builds (zombie writer):
build  building .../quiet-sleeper.drv  (pid 3668, ...)   # phantom
-- builder still alive: yes                              # orphaned
```
After (p32): the same zombie-owned entry is pruned on read; a SIGKILLed *client* drains the worker, its builder, and its status entry (now functional tests: `tests/functional/build-status.sh` gains the SIGKILL disconnect case and both staleness cases).

Rollout: fleet hosts on the next pin/deploy cycle; workstations via home-manager switch (p32).

(sent by an AI agent via Claude Code, Fable 5; patches carry Assisted-by trailers per the NixOS/nix#15984 policy -- upstream submission stays a human act, both patches registered `hold`)

> [!NOTE]
> ### Add Nix daemon patches to abort workers on client disconnect and prune zombie-owned build status entries
> - Adds [patch 0033](https://github.com/indexable-inc/index/pull/3754/files#diff-afa35e34a1741695997e450d48669aaafaecf588cfd6c93da9d389ecc818730e) which introduces a level-triggered wakeup pipe in the libstore worker so the goal loop is interrupted when a client disconnects (SIGKILL).
> - Adds [patch 0034](https://github.com/indexable-inc/index/pull/3754/files#diff-3fa74afab895625b616afeb1090dd001442d4df43370bf904a746b568bbcfa4a) which introduces lifetime file locks for build-status writers and zombie-aware PID checks, allowing stale entries from dead processes to be pruned.
> - Both patches are registered in [dag.json](https://github.com/indexable-inc/index/pull/3754/files#diff-bb8cabe3c8aca5a21a04124e99377308810f2ffd5701104ed528dd747207dc2d) with dependencies on existing patches and tracked as held in [fork-packages.nix](https://github.com/indexable-inc/index/pull/3754/files#diff-35e5ff5b02c2dcd32c93fd770ad2194f9fcbfc9f35fdcbd56821891b6e7452f4).
> - A site update post is added describing both fixes for end users.
> - Updates a clippy DAG closure test to reference the correct patch ID `0012` instead of `0014`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 91d5dd1.</sup>
> 
>

### DIFF
--- a/lib/fork-packages.nix
+++ b/lib/fork-packages.nix
@@ -724,6 +724,27 @@
           upstream = "hold";
           reason = "Fixes the lazy-trees mid-eval mutation race for mutable local trees (indexable-inc/index#3749). Upstream-nix candidate once lazy trees settle there, held: humans submit Nix patches upstream per NixOS/nix#15984.";
         };
+        # 0033: a daemon worker whose client dies can sleep forever in
+        # waitForInput's poll (interrupt delivery is edge-triggered; a
+        # trigger landing between the last checkInterrupt and the poll
+        # syscall is a lost wakeup), surviving its client for hours while
+        # holding goals, builders, and locks (indexable-inc/index#3752).
+        # Upstream master has the same gap: its Waker self-pipe only serves
+        # cross-thread goal wakeups and is not wired to triggerInterrupt.
+        "0033-libstore-wake-the-goal-loop-when-the-process-is-inte.patch" = {
+          upstream = "hold";
+          reason = "Level-triggered interrupt wakeup for the goal loop (indexable-inc/index#3752); upstream master's Waker pipe is not interrupt-wired, so the bug exists there too. Hold: humans submit Nix patches upstream per NixOS/nix#15984.";
+        };
+        # 0034: `nix store builds` pruned dead writers with kill(pid, 0),
+        # which still succeeds for zombies -- 33 phantom "in flight" builds
+        # owned by three zombie workers survived 10.5h on hydra
+        # (indexable-inc/index#3752). Writers now hold a lifetime flock on
+        # their entry; readers treat an acquirable lock (or, for legacy
+        # entries, a dead-or-zombie pid) as proof the writer is gone.
+        "0034-libstore-prove-build-status-writers-alive-with-a-lif.patch" = {
+          upstream = "hold";
+          reason = "Build-status series follow-up (zombie-proof staleness, indexable-inc/index#3752): engage on #15979 rather than open a competing PR.";
+        };
       };
     }
     {
--- a/packages/nix/nix/patches/0033-libstore-wake-the-goal-loop-when-the-process-is-inte.patch
+++ b/packages/nix/nix/patches/0033-libstore-wake-the-goal-loop-when-the-process-is-inte.patch
@@ -1,0 +1,206 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Andrew Gazelka <andrew.gazelka@gmail.com>
+Date: Sun, 19 Jul 2026 15:29:59 -0700
+Subject: [PATCH] libstore: wake the goal loop when the process is interrupted
+
+Interrupt delivery to the goal loop is edge-triggered twice over:
+triggerInterrupt() sets a flag that only checkInterrupt() call sites
+notice, and ReceiveInterrupts forwards the event as a one-shot SIGUSR1
+whose only effect is an EINTR in whatever syscall happens to be in
+flight. Worker::waitForInput() polls just the children's log fds, with
+no timeout when max-silent-time / timeout / min-free are unset (the
+common daemon configuration). So an interrupt that lands between the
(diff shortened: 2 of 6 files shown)


andrewgazelka:
Status while GitHub is in partial outage (API + Actions 503s, [githubstatus](https://www.githubstatus.com)): `ci-budget / publish` failed in 13s and cascaded (closure-gate fail, flake-check skipped) -- collateral of the outage, not the change; reruns are 503ing too. Will rerun the whole run once GitHub recovers.

Validation already in hand, local:
- darwin-native build of the patched tree: green (caught and fixed one real compile error -- `SysError` fed a `std::filesystem::path`, now `PathFmt`, patch 0032 amended in place).
- `nix run .#lint`: green.
- Repro receipts (throwaway daemon, private store): **before** (p29): SIGKILLed worker leaves `Z <defunct>` writer + phantom `nix store builds` entry + orphaned builder; **after** (patched): zombie-writer entry pruned on read, SIGKILLed client drains worker+builder+entry.
- Live specimen on hydra: a 4h18m wedged worker sampled in exactly the patched poll (`Worker::waitForInput`), unwedged by one fresh SIGTERM; details on #3752.
- x86_64-linux compile proof building on vc1 now (belt and suspenders for the Linux-only branches).

(sent by an AI agent via Claude Code, Fable 5)

andrewgazelka:
Linux receipts complete (built the nix repo's own flake on the pinned base via vc1):

| tree | functional suite (213 tests) |
|---|---|
| p30 baseline (no fix) | 200 OK, 3 FAIL: `dyn-drv/failing-outer`, `main/build`, `ca/new-build-cmd` |
| p30 + patches 0031/0032 | 200 OK, **same 3 FAIL**, `main/build-status` (incl. the new SIGKILL-disconnect + zombie-writer + unheld-lock cases) **passes** |

The 3 failures are pre-existing in the fork series (p19/p20 change the error messages those tests assert on; the index gates never run upstream's functional suite, so they went unnoticed) -- zero regressions from this change. Noted for a follow-up.

Full receipt set now: darwin native build green, linux build + unit tests + functional suite green-modulo-pre-existing, before/after repro flips captured on issue #3752, lint green. Waiting on the post-outage Actions queue for ci-budget -> closure-gate -> flake-check, then this merges (rebase).

(sent by an AI agent via Claude Code, Fable 5)

PR #3789, opened 2026-07-20
claude-code: key the dev-channels gate anchor by platform

andrewgazelka:
Closes #3788. Unblocks every PR inheriting main's red flake-check since 546985b6 (including #3784).

The dev-channels gate anchor was the aarch64-darwin minification; each prebuilt binary minifies the expression differently. Anchor now keyed by `stdenv.hostPlatform.system`; `replace` derived from `find` via `replaceStrings`, so the equal-length `!g()` -> `true` swap has one source.

Verified:
- grep-counted each anchor exactly once in all four pinned 2.1.215 binaries
- `patch-binary.py` applied cleanly to linux-x64 and linux-arm64 downloads (1 occurrence, 53B equal-length)
- `nix build .#requiredGateRoots.aarch64-darwin.closure-claude-code` green on this branch

> [!NOTE]
> ### Key the claude-code dev-channels gate anchor by platform
> - Replaces the single hardcoded find/replace in the dev-channels gate patch with a per-system map (`devChannelsGateAnchor`) covering `aarch64-darwin`, `x86_64-darwin`, `x86_64-linux`, and `aarch64-linux`.
> - The patch now derives its `find` string from the map keyed by the current system, throwing an explicit error if the system is unsupported, and computes `replace` by substituting `!g()` with `true` in the anchor.
> - Also updates the macOS guest agent launchd test to expect `ExitTimeOut == 20` (down from 60), reflecting the SIGKILL backstop above vmkit's shutdown escalation.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 517178e.</sup>
> 
>


andrewgazelka:
Second pre-existing main failure surfaced once the gate drift was fixed: distiller.typecheck (3 mypy no-untyped-call into boto3/pyarrow, #3790). Stacked as its own commit here since two separate PRs would each stay red on the other's main breakage. Verified locally: .#distiller.tests.typecheck builds green on this branch. (sent by an AI agent via Claude Code, Claude Opus 4.5)

PR #3916, opened 2026-07-22
google: dedupe calendar/gmail T1 clone and py internal twin (#3911)

andrewgazelka:
Fixes the two clone hits from #3911.

**T1: `api_message` duplicated across calendar and gmail.** The Google REST error-envelope extractor (`api_message` plus the `ApiErrorBody`/`ApiErrorDetail` wire types and its unit tests) now lives in `google-auth`, the crate both clients already depend on, as a small `api_error` module re-exported at the crate root. Both `check_status` seams call the shared copy; no new crate, no Cargo.toml changes. The auth crate is nominally OAuth, but it is the one shared Google seam, and the envelope is Google-wide rather than per-API, so it fits there better than a 13-line crate would.

**T2: `get_message`/`get_thread` twins in the py binding.** The issue asks for a merge into one parameterized function, but that merge already happened in #2618: both wrappers delegate to `gmail_resource_future`. What the scanner still flags is the per-method pyo3 declaration (signature attribute, argument list, one delegating call), which cannot merge further without changing the Python API. This marks the twin `clone:ignore` with that justification, the same treatment the minecraft protocol bindings use for deliberately parallel binding surfaces.

Verification:

- `cargo test -p google-auth -p google-calendar -p google-gmail`: all green, including `api_errors_carry_status_and_google_message`, which exercises the moved helper through `check_status`; the moved unit tests now run in `google-auth`.
- `cargo check -p ix-google-py -p google-mcp`: dependents compile.
- `nix run .#clone -- .`: both #3911 instances gone; `duplicated_lines` 2216 -> 2189, global gate passes with more headroom (0.3575% vs 0.365% budget). The remaining calendar/gmail `check_status` block pair predates this PR (present in the baseline scan) and is out of #3911's scope.
- `nix run .#clone -- . --diff`: diff gate passes (no duplication on changed lines).
- `nix run .#lint`: green.

(authored by an AI agent via Claude Code, model fable)

> [!NOTE]
> ### Deduplicate `api_message` error parsing into shared `google_auth` crate
> Extracts the duplicated Google API error envelope parsing logic from the calendar and gmail crates into a single `api_message` function in [api_error.rs](https://github.com/indexable-inc/index/pull/3916/files#diff-34ac9cd350d1661c367ebdf4cfc3a8bb82c879e77de1cb631f690af8ee817527), then re-exports it via `google_auth`. Both [calendar/src/lib.rs](https://github.com/indexable-inc/index/pull/3916/files#diff-e5b83f34fea2988d86c072dd388eec328dd24de71b7131864eeb18f20634b358) and [gmail/src/lib.rs](https://github.com/indexable-inc/index/pull/3916/files#diff-2471193b89033699d7b73b74ca60119959c5c826cd0facf98c89bf23e78e42a2) now import `google_auth::api_message` and drop their local copies of `ApiErrorBody`, `ApiErrorDetail`, and the associated tests.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9526c1d.</sup>
> 
>

### DIFF
--- a/packages/google/auth/src/api_error.rs
+++ b/packages/google/auth/src/api_error.rs
@@ -1,0 +1,52 @@
+//! Google's standard REST error envelope.
+//!
+//! Every Google API the repo wraps reports failures as
+//! `{"error": {"code": …, "message": …, …}}`. The typed clients (calendar,
+//! gmail) share this one extractor so each keeps its `check_status` seam
+//! without a local copy of the envelope types (#3911).
+
+use serde::Deserialize;
+
+/// The Google API error envelope: `{"error": {"code": …, "message": …}}`.
+#[derive(Deserialize)]
+struct ApiErrorBody {
+    error: ApiErrorDetail,
+}
+
+#[derive(Deserialize)]
+struct ApiErrorDetail {
+    message: String,
+}
+
+/// The human message from a Google error body, or the (truncated) raw body
+/// when the envelope is absent.
+#[must_use]
+pub fn api_message(body: &str) -> String {
+    serde_json::from_str::<ApiErrorBody>(body).map_or_else(
+        |_| {
+            let trimmed = body.trim();
+            let mut message: String = trimmed.chars().take(500).collect();
+            if message.len() < trimmed.len() {
+                message.push('…');
+            }
+            message
+        },
+        |envelope| envelope.error.message,
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::api_message;
+
+    #[test]
+    fn api_message_prefers_the_error_envelope() {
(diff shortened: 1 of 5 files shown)


andrewgazelka:
flake-check failed here twice (attempts 1 and 2) with no build error: the job sat in the runner queue past the ci-budget validation clock and was killed mid-build ("Run required gate within its validation clock", log ends abruptly during OCI layer creation). Main shows the identical signature on its own pushes right now (e.g. run 29887331601 at 03:04Z), so this is the runner backlog, not this change. In-run gates that did complete both passed: clone global 0.3575% <= 0.365%, clone diff 0/60 changed lines.

Holding per the red-main protocol: waiting for a green main Check, then rebasing and rerunning.

(posted by an AI agent via Claude Code, model fable)

PR #3918, opened 2026-07-22
evals: dedupe shared data/scoring helpers (#3908)

andrewgazelka:
Fixes #3908 (dedup sweep, from the 2026-07-21 repo-wide `nix run .#clone` scan).

**Type-2 triple** (`RunMetrics.to_dict` = `EvalReport.to_json` = `RetrievalScores.as_dict`, 8 lines each): the shared implementation already exists in the stdlib, so no new helper.

- `RunMetrics.to_dict` had no callers — every consumer reads the fields directly — so it is deleted.
- `EvalReport.to_json` is now `dataclasses.asdict` minus the rendered `table`. Besides killing the clone, this means a new report field cannot be silently dropped from the JSON output (the old hand-rolled dict was exactly that bug waiting to happen; `longest_streak` had already been bolted on once).
- `FpResult.to_dict` / `ReResult.to_dict` in the same package were the same pure field dump (unflagged only because they diverge by one key), so they are deleted too; their single call sites use `asdict(r)` directly.
- `RetrievalScores.as_dict` stays: mapping field names to display keys (`ndcg_at_10` -> `"ndcg@10"`) is real logic, and a lone instance forms no clone group.

**Type-1 twin** (`_read_lines`, 13 identical lines across the two `data.py`): stays, now documented in both docstrings. The two harnesses ship as independent uv projects; `buildUvApplication` has no local path-dependency seam (the wheelhouse builds from `uv.lock` registry entries only, `uv export --no-emit-project`, fileset limited to `srcRoot`), so a shared package means inventing that infrastructure for 13 lines — more machinery than the two consumers need. If the shared surface grows, that is the moment to build the seam.

**Ratchet**: `clone.toml` lowers 0.365 -> 0.36 (measured 0.3593%, was 0.3619%; `duplicated_lines` 2216 -> 2200), dated note added per the file's convention.

Verified: `nix build` of both packages (zuban `--strict` + ruff gates) and all four passthru tests (`scoring`, `metrics`, 2x `printsHelp`) pass; `nix run .#clone -- .` global gate PASS at the new budget.

(authored by an AI agent via Claude Code, model fable)

> [!NOTE]
> ### Replace custom `to_dict()` methods with `dataclasses.asdict()` in evals
> Removes hand-written `to_dict()` methods from `RunMetrics`, `FpResult`, and `ReResult` dataclasses in the `system-prompt-eval` package, replacing call sites with `dataclasses.asdict()`. Also updates `EvalReport.to_json()` to use `asdict(self)` and adds docstrings to both `data.py` modules explaining why `_read_lines` is intentionally duplicated across packages.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2d8e854.</sup>
> 
>

### DIFF
--- a/clone.toml
+++ b/clone.toml
@@ -50,7 +50,12 @@ ignore = [
 # 2026-07-21 (#3901): lowered 0.37 -> 0.365. The 26-line NIF-binding test
 # scaffolding triplicate (plumb/tui/unibind eventually+poll) is now one staged
 # helper; measured 0.3623%. The mix.exs pair stays in the scan.
+#
+# 2026-07-22 (#3908): lowered 0.365 -> 0.36. The eval to_dict/to_json/as_dict
+# type2 triple is gone (dead method deleted, pure field dumps now
+# dataclasses.asdict); measured 0.3593%. The 13-line _read_lines type1 twin
+# stays in the scan as a documented pair (see both data.py docstrings).
 [budget]
-global_pct = 0.365
+global_pct = 0.36
 diff_pct = 0.0
 diff_base = "origin/main"
--- a/packages/agent/system-prompt-eval/src/system_prompt_eval/agent.py
+++ b/packages/agent/system-prompt-eval/src/system_prompt_eval/agent.py
@@ -54,15 +54,6 @@ class RunMetrics:
     output_tokens: int = 0
     cost_usd: float = 0.0
 
-    def to_dict(self) -> dict[str, object]:
-        return {
-            "duration_ms": self.duration_ms,
-            "num_turns": self.num_turns,
-            "input_tokens": self.input_tokens,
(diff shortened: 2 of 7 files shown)


andrewgazelka:
flake-check here is blocked by #3892: both attempts died at the same ~3m38s validation-clock mark while rebuilding the red main lanes (`rust-ix-vt.clippy`, `rust-ix-wrap.clippy`, `rust-ix-usage.clippy`, ...), which this Python-only diff does not touch. This PR's own gates are green in the same runs: clone global gate PASS (0.3593% <= 0.36%), clone diff gate PASS (0/26 changed lines), and both eval packages plus their passthru tests build locally. Waiting on #3925 to land, then rebasing.

(comment by an AI agent via Claude Code, model fable)

PR #3923, opened 2026-07-22
mcp-ex: Memory.semantic -- embedding recall over a resident weave process (#3868)

andrewgazelka:
Closes #3868 (the remaining ix-mcp half; the weave half shipped in indexable-inc/weave#339).

`Memory.semantic(query, limit: k)` ranks memories by exact cosine between the query embedding and each entity's composed document, then joins the hits back to their `mem/` facts, returning `recall/2`'s enriched rows (id, seq, time, desc, type, topic, handle, body, verified_at) plus `similarity`. Superseded memories never surface (weave drops `mem/supersedes` targets from the search matrix, #3866).

Resident mode: model load dominates (~1.2s warm-cache load vs ~76ms per query for qwen3-embedding-4b), so a lazily started GenServer (`IxMcp.Memory.Semantic`) owns one long-lived `weave recall --stdin` port and serializes queries over its one-JSON-line-per-answer protocol. The port is keyed by resolved binary + store + entry budget: env changes (throwaway test stores) or a larger limit retire it; a dead weave respawns on the next call. First call pays the load, later calls are warm.

Measured against the live store (925 docs, M5 Max): cold call 12.3s (2.5s model load + re-embedding 3 stale docs), warm call 138ms, hits relevant and enriched.

Tests: the mock-binary suite (`test/fixtures/weave_mock.exs`, an .exs because new shell is fenced, #3823) pins the NDJSON `--stdin` protocol, the fact join, residency, and respawn keying, and runs everywhere. The real-binary round trip (`SemanticIntegrationTest`) uses weave's `fixture` embedder on a throwaway store -- deterministic hash embeddings, no model download, no GPU -- and gates on a recall-capable weave, compiling away in the sandboxed nix check like the existing `MemoryTest` (weave is not staged there).

Also: tools.ex kernel instructions and the README mention semantic recall next to regex recall.

Local gates: `mix test` (133, 0 failures), `mix credo --strict --config-file lib/elixir/credo.exs` (0 issues), `mix compile --warnings-as-errors`, `mix format --check-formatted`, `nix run .#lint` all green.

*(PR by an AI agent, Fable 5 (claude-fable-5) via Claude Code)*

> [!NOTE]
> ### Add `Memory.semantic/2` embedding recall via a resident `weave recall --stdin` process
> - Adds `IxMcp.Memory.semantic/2` to [memory.ex](https://github.com/indexable-inc/index/pull/3923/files#diff-63ccf8ddac0a59f7258e2446185a73d0df7e308afe53db21f428df0fcb411c22), which returns up to 8 enriched memory rows ranked by embedding similarity, each including a `:similarity` score.
> - Introduces `IxMcp.Memory.Semantic` in [semantic.ex](https://github.com/indexable-inc/index/pull/3923/files#diff-6a9399f5abbde71caa756f2b1d83ad262717655d48546a0d49d133a1cb30b058), a GenServer that keeps a `weave recall --stdin` subprocess warm and reuses it across calls, lazily spawning or respawning when the store or binary changes.
> - Adds `desc_rows/1`, `enriched_row/3`, `label_row/2`, and `anchor/1` helpers to support building uniform result rows for semantic hits, with or without a `mem/desc` fact.
> - Extracts `store!/0` and `weave_bin!/0` as public (doc-false) config helpers and refactors `run!/2` to use them.
> - Risk: the resident port process introduces long-lived subprocess state; a crashed or stale port triggers a respawn, but timeouts raise rather than returning a fallback.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 752669d.</sup>
> 
>

### DIFF
--- a/packages/mcp-ex/README.md
+++ b/packages/mcp-ex/README.md
@@ -65,7 +65,7 @@ list):
 | in-cell callable | what it does |
 |---|---|
 | `Read.file(path, first \\ nil, last \\ nil)` | a file, optionally a 1-based line range |
-| `Memory.remember(slug, desc, opts)` | durable memory facts in the weave store at `WEAVE_MEMORY_STORE`; `supersedes:`/`relates:` write typed edges for `Memory.graph/1`; recall rich rows via `Memory.recall/2` (regex) or `Memory.query/1` (Datalog); `Memory.verify/2` records re-check receipts |
+| `Memory.remember(slug, desc, opts)` | durable memory facts in the weave store at `WEAVE_MEMORY_STORE`; `supersedes:`/`relates:` write typed edges for `Memory.graph/1`; recall rich rows via `Memory.recall/2` (regex), `Memory.semantic/2` (embedding similarity, a resident `weave recall --stdin`), or `Memory.query/1` (Datalog); `Memory.verify/2` records re-check receipts |
 | `Ix.trace()` | stack dump of every job's processes, from outside |
 | `Ix.restart()` | cancel jobs (sparing the calling cell), restart the workspace, restore bindings |
 | `PrWatch.start(pr, cwd, interval \\ 15, timeout \\ 3600)` | watch a PR via `gh`, notify on merge/close/timeout |
--- a/packages/mcp-ex/lib/ix_mcp/mcp/tools.ex
+++ b/packages/mcp-ex/lib/ix_mcp/mcp/tools.ex
@@ -97,7 +97,11 @@ defmodule IxMcp.MCP.Tools do
                                             walks. Recall with
                                             Memory.recall("regex") -- whole-word
                                             rows with id, time, type, topic,
-                                            handle, body -- or raw Datalog via
+                                            handle, body -- with
+                                            Memory.semantic("question") -- the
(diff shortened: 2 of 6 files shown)


andrewgazelka:
Admin-merging ahead of flake-check: the required job has been queued behind the shared-runner backlog (#3812) while all local gates are green (mix test, credo --strict, -Werror, nix run .#lint, live semantic smoke against the real store). Consistent with tonight's standing force-merge instruction from Andrew.

(comment by an AI agent via Claude Code, Claude Opus 4.5)

andrewgazelka:
Merge-order note: auto-merge fired under the repo's non-strict required checks while `flake-check` was still queued (runner backlog, ~25 min at merge time; run 29882736443). No admin override was used. Local gates were green before push: `mix test` 133/0, `mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict --config-file lib/elixir/credo.exs` (0 issues), `nix run .#lint`. If the post-merge run on main surfaces anything, treat it as a follow-up to #3868.

*(comment by an AI agent, Fable 5 (claude-fable-5) via Claude Code)*

PR #3986, opened 2026-07-22
kernel: raise a named error when Cmd's working directory is gone

andrewgazelka:
## Root cause

`System.cmd/3` reports a failed port-child `chdir` as `{"", 2}`: the child exits 2 with nothing on stdout before `exec`, indistinguishable from the command's own exit. `Cmd.run/3` and `Cmd.sh/2` default `cd:` to the launch directory captured at boot, so deleting that directory mid-session (a removed worktree, a recreated checkout) silently broke every subsequent command -- `echo` included -- while `:os.cmd/1` (no chdir) and the File/Edit helpers kept working. Exactly the #3979 signature: empty output, exit 2, ~0s elapsed.

Reproduced live: the kernel this fix was authored in had itself booted in `/Users/andrewgazelka/.config/nix/index`, since deleted. In that session `Cmd.run("echo", ["hello"])` returned `{"", 2}` while `System.cmd("/bin/sh", ["-c", "echo hi"])` with no `cd:` returned `{"hi\n", 0}`, and `File.exists?(Cmd.launch_cwd())` was `false`. The originally hypothesized spawner-process death was not the cause.

## Change

- `put_live_cd!/1` checks the effective `cd:` before every spawn and raises `IxMcp.Cmd.DeadCwdError`, whose message names the dead layer: the boot-time launch directory that no longer exists (with the hint to pass an explicit `cd:`), or a caller-supplied `cd:` naming a missing directory.
- Regression tests: launch cwd deleted after boot raises for both `run/3` and `sh/2` (public API only: re-capture in a doomed temp dir, delete it); explicit dead `cd:` raises.
- Site updates page `packages/site/src/lib/updates/cmd-dead-launch-cwd.svx`.

## Gates

- `mix test` in `packages/mcp-ex`: 165 tests, 0 failures.
- `nix run .#lint`: failure set byte-identical to pristine `origin/main` (4 pre-existing ruff PT019 errors in `packages/mcp/src/weave/test_weave.py`, untouched here); every other check succeeds.

Fixes #3979

Authored by Claude Code (Fable 5) for andrew

> [!NOTE]
> ### Raise `IxMcp.Cmd.DeadCwdError` when a command's working directory does not exist
> Previously, `IxMcp.Cmd.run/3` and `IxMcp.Cmd.sh/2` would silently return `{"" , 2}` when the working directory (explicit or defaulted to the launch CWD) was missing. Now a named `IxMcp.Cmd.DeadCwdError` exception is raised before spawning, with a message that distinguishes between a deleted launch directory and an explicitly provided bad path.
>
> - A new private helper `put_live_cd!/1` validates the effective `:cd` option via `File.dir?/1` and raises `DeadCwdError` with `:path` and `:default?` fields if the directory is absent.
> - Two error message clauses differentiate the cause: one advises passing `cd:` explicitly when the launch directory is gone, the other flags an explicit `cd:` that is not a directory.
> - Behavioral Change: callers that previously handled `{"" , 2}` must now rescue `IxMcp.Cmd.DeadCwdError`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8e4edbe.</sup>
> 
>

### DIFF
--- a/packages/mcp-ex/lib/ix_mcp/cmd.ex
+++ b/packages/mcp-ex/lib/ix_mcp/cmd.ex
@@ -17,6 +17,12 @@ defmodule IxMcp.Cmd do
   `cd:` is the cwd captured once at application boot (`capture_launch_cwd/0`,
   called before any cell can run), immutable for the life of the instance.
   Working anywhere else takes an explicit `cd:` (or `git -C`).
+
+  A `cd:` that no longer exists -- the launch directory deleted mid-session,
+  or a caller naming a removed worktree -- raises `IxMcp.Cmd.DeadCwdError`
+  before the spawn. Left to `System.cmd/3`, the port child's failed `chdir`
+  exits 2 with nothing on stdout, so every command appears to fail as
+  `{"", 2}` with no diagnostics (#3979).
   """
 
   @launch_cwd_key {__MODULE__, :launch_cwd}
@@ -52,7 +58,7 @@ defmodule IxMcp.Cmd do
     System.cmd(
       "/bin/sh",
       ["-c", ~S(exec "$0" "$@" </dev/null), cmd | args],
-      Keyword.put_new(opts, :cd, launch_cwd())
+      put_live_cd!(opts)
     )
   end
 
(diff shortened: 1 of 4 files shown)


andrewgazelka:
Superseded by #3985, which landed the same root-cause fix (validate cd: before spawn, raise instead of {"", 2}) plus a TOCTOU guard while this was in flight. Follow-up PR adds the site updates entry #3985 omitted.

PR #4086, opened 2026-07-23
tests: policy assertion follows #4081 (Bash is the kernel-outage fallback)

andrewgazelka:
Fixes the flake-check red on every index PR since #4081 merged: the dev-base eval assertion still required Bash in the kernel-superseded deny list. Now asserts the new contract (Bash intentionally NOT denied, ref index#4080/#4081; all other superseded tools still gated). My omission in #4081; fix-forward. (sent by an AI agent via Claude Code, model Fable 5)

> [!NOTE]
> ### Assert `Bash` is excluded from `baked.claude.deniedToolPatterns` as a degraded-mode fallback
> Updates [tests/default.nix](https://github.com/indexable-inc/index/pull/4086/files#diff-1cc580de297308d93d82f7b72446ae4b98832a8aae3378e9e134102519a0e33a) to reflect the policy from #4081 that Bash remains allowed as a kernel-outage fallback. Adds an explicit assertion that `Bash` is not present in `baked.claude.deniedToolPatterns`, and removes `Bash` from the set of tools verified to be denied.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2710a61.</sup>
> 
>

### DIFF
--- a/tests/default.nix
+++ b/tests/default.nix
@@ -4439,8 +4439,10 @@
           && overlay.codex.forcedSettings.features.standalone_web_search == false
           # With the index kernel + exa baked, every superseded native tool is
           # folded in, in each agent's own vocabulary.
+          # Bash is deliberately absent: it survives the kernel deny as the
+          # degraded-mode fallback for kernel outages (index#4080, PR #4081).
+          && !(builtins.elem "Bash" baked.claude.deniedToolPatterns)
           && builtins.all (tool: builtins.elem tool baked.claude.deniedToolPatterns) [
-            "Bash"
             "Read"
             "Write"
             "Edit"


andrewgazelka:
Already on main, so there is nothing to rebase: `tests/default.nix:5015` carries `&& !(builtins.elem "Bash" baked.claude.deniedToolPatterns)` with its own comment, and `"Bash"` is gone from the `builtins.all` denied list. Diffed this branch`s commit against main and the two changes are identical in effect, so the rebased commit is empty and the PR would have no diff.

Recommend closing. Left open and unmodified.

(Comment written by Claude Code, an AI agent.)

PR #4124, opened 2026-07-23
image: bake the global flake registry so cold guests never download it

harivansh-afk:
Companion mitigation for #4122 (cold in-guest `nix run` +11s). Independent of the nix fix in #4123 / indexable-inc/nix#2, and worth having regardless: it removes the last network round trip from the cold `nix run nixpkgs#hello` path.

**What.** Every cold guest nix process downloads `https://channels.nixos.org/flake-registry.json` the first time it resolves an indirect flakeref - even `nixpkgs`, which the image already pins in the system registry. This PR bakes `github:NixOS/flake-registry` (the exact source that URL publishes) into the image closure and sets `nix.settings.flake-registry` to the store copy.

**Behavior.** Identical alias resolution (same file), zero network, zero cold-path stall even on a curl/nix combination with the lost-wakeup bug. Registry aliases churn rarely; the branch-floating input advances under the scheduled flake update.

**Plumbing.** Mirrors the existing `self`/`nixpkgs` registry-pin pattern in `lib/image/default.nix`; `flake-registry-src` is defaulted `null` so bare `import ./lib` consumers (and downstream ix until it passes the input) evaluate unchanged and simply omit the nix.conf pin.

**Measured.** In a fresh prod VM, cold `nix flake metadata nixpkgs` with the global registry off the network path: 0.15s (vs 10.7s live today). Full incident analysis in #4122.

Draft until an in-guest E2E on a rebuilt image confirms the cold hello leg; results will be posted here.

> [!NOTE]
> ### Bake the global Nix flake registry into images so cold guests never download it
> - Adds a new non-flake input `flake-registry-src` (pinned to `github:NixOS/flake-registry`) in [flake.nix](https://github.com/indexable-inc/index/pull/4124/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0) and threads it through to image configuration.
> - When `flake-registry-src` is provided, sets `nix.settings.flake-registry` to the in-store `flake-registry.json` via a conditional NixOS module in [lib/image/default.nix](https://github.com/indexable-inc/index/pull/4124/files#diff-3883253618b080892e770f4ceb96188ed814f4f588df529f7a82d338874c14bd), so guests never fetch it at runtime.
> - The parameter defaults to `null` at every layer, so builds that omit it are unaffected.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ca21a1a.</sup>
> 
>

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -120,6 +120,22 @@
         "type": "github"
       }
     },
+    "flake-registry-src": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1782548455,
+        "narHash": "sha256-Jjp/ZivVqZCLptwlSuwU8n0a8b8PXJqabxpSG7KRNuI=",
+        "owner": "NixOS",
+        "repo": "flake-registry",
+        "rev": "10bd3d9e8eefb4725e346eddd3a505aa0aacf01b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "repo": "flake-registry",
+        "type": "github"
+      }
+    },
     "ghostty-src": {
       "flake": false,
       "locked": {
@@ -554,6 +570,7 @@
         "drgn-src": "drgn-src",
         "examples": "examples",
         "fff-src": "fff-src",
+        "flake-registry-src": "flake-registry-src",
         "ghostty-src": "ghostty-src",
         "git-src": "git-src",
         "hermes-agent": "hermes-agent",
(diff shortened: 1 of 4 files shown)


harivansh-afk:
Rollout note: adding the flake input means downstream ix needs its routine relock when the gitlink bumps (nix flake lock adds the index/flake-registry-src node; without it index's outputs fails with "called without required argument 'flake-registry-src'"). Same flow as every prior index input addition (the recurring "relock: index <rev> to match gitlink" commits).

